PR #14, opened 2016-11-16
fix(style-strategy): validate method returns correct validation message

eriklieben:
Noticed a typo/ incorrect error message while reading through the code / trying to understand it's inner workings.

Don't know the best way to report this, so just created a pull.


eriklieben:
I should drink coffee before doing something in the morning, pull request was already created :-)

PR #23, opened 2016-12-01
feat(tooltip): add ux-tooltip component

ctjhoa:
Usage:
```html
        <ux-tooltip>
          <div slot="tooltip-target">
            <ux-button>Button</ux-button>
          </div>
          <div slot="tooltip-content">
            Tooltip
          </div>
        </ux-tooltip>
```


src/tooltip/ux-tooltip.ts
```diff
@@ -0,0 +1,32 @@
+import {customElement, bindable, ViewResources, View, processAttributes} from 'aurelia-templating';
+import {inject} from 'aurelia-dependency-injection';
+import {StyleEngine} from '../styles/style-engine';
+import {Themable} from '../styles/themable';
+import {processDesignAttributes} from '../designs/design-attributes';
+
+@inject(ViewResources, StyleEngine)
+@customElement('ux-tooltip')
+@processAttributes(processDesignAttributes)
+export class UxButton implements Themable {
```
stephenlautier: UxTooltip? :)

EisenbergEffect:
@ctjhoa Thanks for working on this. I'm wondering if we could do something a bit different with the API. For example, what about this:

```html
<ux-button ref="myButton">Click Me</ux-button>

<ux-tooltip target.bind="myButton">
  The tooltip content goes here.
</ux-tooltip>
```

I'm also wondering if we could have a simpler usage for tooltips that are always plain text. Perhaps something like this:

```html
<ux-button ux-tooltip="The tooltip content goes here.">Click Me</ux-button>
```

So, you could use the custom attribute for the simplest, most common scenario and then use the element for the more complex scenarios, with a target binding to the element it is attached to.

Thoughts?

tochoromero:
How do you have both a Custom Element and a Custom Attribute with the same name? Would Aurelia be smart enough to know you are using one or the other? What about import conflicts?

ctjhoa:
@EisenbergEffect I play with css positioning to place the tooltip under the target without javascript computing. So I need a "wrapper" in `position: relative` in order to apply a `position: absolute` on the tooltip.
I don't see how to position the tooltip in your example:
```html
<ux-button ref="myButton">Click Me</ux-button>

<ux-tooltip target.bind="myButton">
  The tooltip content goes here.
</ux-tooltip>
```
Any idea?

moodledan:
Something like this perhaps: http://jsfiddle.net/RUjBh/6/

ctjhoa:
@moodledan I don't see your point. Your fiddle solve the second example:
```html
<ux-button ux-tooltip="The tooltip content goes here.">Click Me</ux-button>
```
not the first one as you work only with the node `.tooltip`. Or I miss something.
My point is: I don't know how I can, with just css, position tooltip on a node from a reference.

moodledan:
Apologise @ctjhoa I misread/speedread and missed your point, how about this 

[Position an element relatively to another element from anywhere in the DOM](https://discourse.wicg.io/t/position-an-element-relatively-to-another-element-from-anywhere-in-the-dom/968)

ctjhoa:
@moodledan This describe exactly our problem but this is a spec draft.
Only [the lastest post](https://discourse.wicg.io/t/position-an-element-relatively-to-another-element-from-anywhere-in-the-dom/968/49) seems to show a solution available today but it require http://elementqueries.com/ and I'm not sure that's a dependency we want (regarding the youth/stars of the project).

The more I think about it, the more I realize that's a limitation of component architecture. The isolation that brings component require also style isolation. This philosophy is opposed to the css styling because in css you can style element depending of its sibling/parent or completely different node.

EisenbergEffect:
I'm going to close this PR for now since it's not quite the direction we want to go.

PR #26, opened 2016-12-06
Material Design shadows

moodledan:
Added box-shadow shadow/elevation design resources, updated the material-design and ios-design themes to include references to these new resources so they can be applied using the $design. notation in component themes. 

Updated the ux-button theme to use the new shadow styles and added a new style to the button for the :focus state.

### DIFF
--- a/src/button/ux-button-theme.css
+++ b/src/button/ux-button-theme.css
@@ -44,7 +44,7 @@ styles.button.raised, styles.button.fab {
 styles.button.raised {
   transition: box-shadow 0.2s cubic-bezier(0.4, 0, 0.2, 1);
   transition-delay: 0.2s;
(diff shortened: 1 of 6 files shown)


ctjhoa:
I'm not a reviewer but your commits need to match the [Git Commit Guidelines](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit) as specified in [CONTRIBUTING.md](https://github.com/aurelia/ux/blob/master/CONTRIBUTING.md)

EisenbergEffect:
I like this idea. I need to ponder the naming. One thing we'll need to change is the use of underscores, since that's not common in Aurelia's naming patterns. We might want to change a few other things. Overall, I think this will be a nice addition.

PR #27, opened 2016-12-08
Loading components - progress bar and spinner

moodledan:
Material design loading components:

- Determinate and indeterminate progress bar
- Spinner in 3 sizes; small, medium and large

### DIFF
--- a/src/progress/ux-progress-theme.css
+++ b/src/progress/ux-progress-theme.css
@@ -1,0 +1,106 @@
+styles.progress {
+  position: relative;
+  height: 4px;
+  display: block;
+  width: 100%;
+  background-color: ${$design.accentLight};
+  overflow: hidden;
+}
+
+styles.progress .bar {
+  background-color: ${$design.accent};
+  transition: width .3s linear;
+}
+
+styles.progress.progressDeterminate .bar {
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+}
+
+styles.progress.progressIndeterminate .bar:before {
+  content: '';
+  position: absolute;
+  background-color: inherit;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  will-change: left, right;
+  -webkit-animation: indeterminate 2.1s cubic-bezier(0.65, 0.815, 0.735, 0.395) infinite;
+  animation: indeterminate 2.1s cubic-bezier(0.65, 0.815, 0.735, 0.395) infinite;
+}
+
+styles.progress.progressIndeterminate .bar:after {
+  content: '';
+  position: absolute;
+  background-color: inherit;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  will-change: left, right;
+  -webkit-animation: indeterminate-short 2.1s cubic-bezier(0.165, 0.84, 0.44, 1) infinite;
+  animation: indeterminate-short 2.1s cubic-bezier(0.165, 0.84, 0.44, 1) infinite;
+  -webkit-animation-delay: 1.15s;
+  animation-delay: 1.15s;
+}
+
+@-webkit-keyframes indeterminate {
+  0% {
+    left: -35%;
+    right: 100%;
+  }
+  60% {
+    left: 100%;
+    right: -90%;
+  }
+  100% {
+    left: 100%;
+    right: -90%;
+  }
+}
+
+@keyframes indeterminate {
+  0% {
+    left: -35%;
+    right: 100%;
+  }
+  60% {
+    left: 100%;
+    right: -90%;
+  }
+  100% {
+    left: 100%;
+    right: -90%;
+  }
+}
+
+@-webkit-keyframes indeterminate-short {
+  0% {
+    left: -200%;
+    right: 100%;
+  }
+  60% {
+    left: 107%;
+    right: -8%;
+  }
+  100% {
+    left: 107%;
+    right: -8%;
+  }
+}
+
+@keyframes indeterminate-short {
+  0% {
+    left: -200%;
+    right: 100%;
+  }
+  60% {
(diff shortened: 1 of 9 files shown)


moodledan:
Progress bar probably needs a code review, not sure setting the width of the bar in the way I have is correct?

`ux-progress.ts`

```
@bindable public barWidth = {
    width: '60%',
};
```

Then in the template:

```
...
<div class="bar" css.bind="barWidth"></div>
...
```

I have a basic example of the 2 progress bars and spinner to submit to the [app-ux-showcase](https://github.com/aurelia/app-ux-showcase)

moodledan:
Example usage here: https://github.com/moodledan/app-ux-showcase/tree/moodledan-loading-control-examples/src/components

ctjhoa:
I'm not a reviewer but your commits need to match the [Git Commit Guidelines](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit) as specified in [CONTRIBUTING.md](https://github.com/aurelia/ux/blob/master/CONTRIBUTING.md)

EisenbergEffect:
@moodledan For these sorts of scenarios, rather than using top, left, width, height, etc. we'll want to use CSS Transform properties. The reason for this is that the transform properties are hardware accelerated and will perform and animate smoother, especially on mobile devices.

moodledan:
Is there a possibility that effort is being duplicated creating components in the material design style when a [material-components](https://github.com/material-components/) exist in Web, iOS and Android flavours?  

The Web components repo also has an aurelia sample: https://github.com/material-components/material-components-web/tree/71f6162747380a11034f000a5eeba116c8b2136a/framework-examples/aurelia

EisenbergEffect:
These components aren't intended to be only material design, they are also intended to be iOS and eventually Windows. Same API, design language changing based on underlying platform.

moodledan:
Is the thinking that just CSS/Styles will change per platform or is there going to be more granularity. For example the ripple effect on buttons is material but isn't part of iOS and Windows design language, will there be a mechanism to disable/remove the ripple behaviour based on platform?

EisenbergEffect:
That's the idea, eventually.

src/progress/ux-progress.ts
```diff
@@ -0,0 +1,35 @@
...
+import {Themable} from '../styles/themable';
+import {processDesignAttributes} from '../designs/design-attributes';
+
+@inject(ViewResources, StyleEngine)
+@customElement('ux-progress')
+@processAttributes(processDesignAttributes)
+export class UxProgress implements Themable {
+  @bindable public type = null;
+  @bindable public theme = null;
+  @bindable public barWidth = {
```
ZHollingshead: I would very much like to see the width of the bar bound to `value` which will align with the standard `<progress>` HTML5 tag. The `progress` tag has two attributes: `value` which represents the current progress, and `max` which indicates the value that represents full.

`max` could have a default value of 100 so that it is not required. Then we can just calculate the percentage based on `value` and `max`. You could use a Computed Property to keep the HTML how it is.

Example
``` javascript
@computedFrom('this.max', 'this.value')
public get barWidth(){
    return `width: ${(this.value / this.max) * 100}%`;
}
```

EisenbergEffect:
I haven't looked at the latest, but let's make sure that we use animations that are based on transforms instead of width, if at all possible. The transforms will be hardware accelerated on mobile which will make big difference in performance.

src/progress/ux-progress-theme.css
```diff
@@ -0,0 +1,106 @@
...
+  left: 0;
+  bottom: 0;
+  will-change: left, right;
+  -webkit-animation: indeterminate-short 2.1s cubic-bezier(0.165, 0.84, 0.44, 1) infinite;
+  animation: indeterminate-short 2.1s cubic-bezier(0.165, 0.84, 0.44, 1) infinite;
+  -webkit-animation-delay: 1.15s;
+  animation-delay: 1.15s;
+}
+
+@-webkit-keyframes indeterminate {
```
ZHollingshead: Needs to use `transform` instead of `left` or `right`

src/progress/ux-progress-theme.css
```diff
@@ -0,0 +1,106 @@
...
+    left: 100%;
+    right: -90%;
+  }
+  100% {
+    left: 100%;
+    right: -90%;
+  }
+}
+
+@keyframes indeterminate {
```
ZHollingshead: Needs to use `transform` instead of `left` or `right`

src/progress/ux-progress-theme.css
```diff
@@ -0,0 +1,106 @@
...
+    left: 100%;
+    right: -90%;
+  }
+  100% {
+    left: 100%;
+    right: -90%;
+  }
+}
+
+@-webkit-keyframes indeterminate-short {
```
ZHollingshead: Needs to use `transform` instead of `left` or `right`

src/progress/ux-progress-theme.css
```diff
@@ -0,0 +1,106 @@
...
+    left: 107%;
+    right: -8%;
+  }
+  100% {
+    left: 107%;
+    right: -8%;
+  }
+}
+
+@keyframes indeterminate-short {
```
ZHollingshead: Needs to use `transform` instead of `left` or `right`

ZHollingshead:
@moodledan have you had a chance to look at the proposed changes?

moodledan:
Sorry @ZHollingshead I've not had chance to review or re-work the control yet.

ZHollingshead:
It is quite alright, I was just wanting to see where everything stood.

EisenbergEffect:
Closing this for now since there are conflicts and the control needed to be re-worked. Please submit a new PR when ready. Thanks!

Issue #30, opened 2017-01-04
Build fails


moodledan:
**I'm submitting a bug report**

* **Library Version:**
aurelia-ux 0.1.14

**Please tell us about your environment:**
* **Operating System:**
OSX 10.11.5

* **Node Version:**
7.2.1

* **NPM Version:**
3.10.10

* **JSPM OR Webpack AND Version**
JSPM 0.16.48

* **Browser:**
N/A

* **Language:**
TypeScript 2.1.4 

**Current behavior:**
When running `npm run build` the following error is thrown, `npm install` and `jspm install` were run before attempting to run the build.

`$ npm run build`

```
> aurelia-ux@0.1.14 prebuild /Users/dan/Documents/spikes/aurelia-ux/ux-master
> cross-env rimraf dist

```

```
> aurelia-ux@0.1.14 build /Users/dan/Documents/spikes/aurelia-ux/ux-master
> concurrently "npm run build:amd" "npm run build:commonjs" "npm run build:es2015" "npm run build:system"
```

```
[0] 
[0] > aurelia-ux@0.1.14 prebuild:amd /Users/dan/Documents/spikes/aurelia-ux/ux-master
[0] > cross-env rimraf dist/amd
[0] 
```
```
[1] 
[1] > aurelia-ux@0.1.14 prebuild:commonjs /Users/dan/Documents/spikes/aurelia-ux/ux-master
[1] > cross-env rimraf dist/commonjs
[1] 
```
```
[2] 
[2] > aurelia-ux@0.1.14 prebuild:es2015 /Users/dan/Documents/spikes/aurelia-ux/ux-master
[2] > cross-env rimraf dist/es2015
[2] 
```
```
[3] 
[3] > aurelia-ux@0.1.14 prebuild:system /Users/dan/Documents/spikes/aurelia-ux/ux-master
[3] > cross-env rimraf dist/system
[3] 
```
```
[0] 
[0] > aurelia-ux@0.1.14 build:amd /Users/dan/Documents/spikes/aurelia-ux/ux-master
[0] > cross-env tsc --project tsconfig.build.json --outDir dist/amd --module amd
[0] 
```
```
[1] 
[1] > aurelia-ux@0.1.14 build:commonjs /Users/dan/Documents/spikes/aurelia-ux/ux-master
[1] > cross-env tsc --project tsconfig.build.json --outDir dist/commonjs --module commonjs
[1] 
```
```
[2] 
[...]

PR #31, opened 2017-01-09
Build fails #30 - quick fix

verbedr:
This fixes the build but to do so I have set the "noImplicitAny" to false in the tsconfig.json. I'm not sure what the team take on this is. I assume it needs to stay true but then someone need to check out what to do with those implicit any's ....

btw I'm not sure that the doc/tsconfig.json is still used.


EisenbergEffect:
@jdanyow Can you take a look at this and provide some commentary on the TS settings?

verbedr:
@jdanyow the setting that I changed was noImplicitAny. Because the any keyword was used in the code in places that I could not quickly fix, so I have set it to false. I do not know if there is a general set of TS Settings at Aurelia we need to respect. Or that you would object the use of the any keyword in general. If so let me know.

verbedr:
@EisenbergEffect I reverted back the TS settings. Apparently this was caused by a bug in TS 2.1.4. After upgrading the build works with the option enabled.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "module": "amd",
     "moduleResolution": "node",
-    "target": "es5",
+    "target": "es6",
```
ZHollingshead: What does changing the target from es5 to es6 do specifically? I did not notice any big differences but I may have overlooked something.

verbedr: Zach,
I'm not sure any more this has been a while ago. It could be that I was changing things by trial and error. 

As far as I can remember there were 2 different tsconfig specified by in the project. One in the root and one in the doc folder (which was explicitly used by the document generation step in the build). If I remember correct generating the documentation was also causing some errors. So I merged these settings together (opting for the most modern setting) and pointed the doc generation step to the root tsconfig.

Looking up the meaning of the settings in https://www.typescriptlang.org/docs/handbook/compiler-options.html. It doesn't reveal to much but I guess it will be the output version of the JavaScript generated by the tsc. Checking the http://caniuse.com/#search=es6 this means that IE11 is not supported.

I guess it would be better to reset it back to ES5?

To make your decision a bit harder ;-) I added an issue link to another aurelia project (bootstrap) which is using the "es2015" (which is the official es6 name). https://github.com/aurelia/bootstrapper/issues/42. We could opt to align this setting with other au projects?

Let me know what you want it to be.  

Kind Regards
Dries

ZHollingshead: That was my best guess as well. For compatibility reasons, I would want to keep it at ES5 since this change would isolate anyone who is not running windows 10 and uses IE. Users in many enterprise environments for example.

verbedr: PR is up to date and updated with the remark.

ZHollingshead:
@EisenbergEffect What do you think? This resolves the failing build issue on my end.

EisenbergEffect:
@ZHollingshead There seem to be other changes here mixed in with the build fix. Do those relate somehow?

@jdanyow Can you review/comment on the build change? Any issues there?

ZHollingshead:
@EisenbergEffect The changes to paper-wave.ts, style-controller.ts, and style-resource.ts all resolve typescript errors that are thrown during the build process causing it to fail. This happens in my environment as well. #30 has a stack trace detailing one of those errors.

khuongduybui:
I got the same errors with tsc as well.

EisenbergEffect:
Ok, cool. @jdanyow can you review the build config change? Was there some reason we were targeting es6? Was that an accident?

StrahilKazlachev:
@EisenbergEffect If you were looking at the build scripts, the ES6 target is for generating the source of the API doc, not the dist.

jdanyow:
I think ux is out of sync with validation in terms of the package.json's `scripts` and `devDependencies` sections. I would try the following:

1. deleting your ux project's node_modules folder,
2. copying in the `scripts` and `devDependencies` sections from aurelia validation,
3. npm install,
4. `npm run lint`, then fix all the lint issues (the rules change from time to time as TypeScript gets new features).
5. `npm run build` (just to verify it builds)
6. `npm run test`
7. `npm run doc` (just to verify the docs build)
8. update this PR

jdanyow:
superceded by https://github.com/aurelia/ux/pull/42

Issue #25, opened 2016-12-05
UX-Button Icon Type


khuongduybui:
Simply ```<ux-button type="flat fab"/>``` should give me a round (fab) button with no box shadow (flat).
This can be used for [mini secondary actions inside a list](https://material.google.com/components/lists.html#lists-actions).

Here's how I implemented it in my own project (which consumes aurelia/ux):
```
ux-button
  button.flat.fab
    margin 0
    padding 0
    min-width auto
    box-shadow none
```

moodledan:
Material Design has guidelines for the use of [shadows/elevation](https://material.google.com/material-design/elevation-shadows.html#) so I'd be tempted to include it (not include it) in the same way as swatches. 

Pseudo code used below to demonstrate: 

Perhaps a `shadows.ts` containing:

```
export const shadows = {
  depth_0: ' none',
  depth_1: '0 2px 2px 0 rgba(0,0,0,0.14),0 1px 5px 0 rgba(0,0,0,0.12),0 3px 1px -2px rgba(0,0,0,0.2)',
  depth_2: '0 4px 5px 0 rgba(0,0,0,0.14),0 1px 10px 0 rgba(0,0,0,0.12),0 2px 4px -1px rgba(0,0,0,0.3)',
  depth_3: etc ...
};
```

Then update `material-design.ts` with:

```
import {Design} from './design';
import {swatches} from '../colors/swatches';
import {shadows} from '../colors/shadows';

export class MaterialDesign implements Design {
  public type = 'material';

  public shadow_elevation_none = shadows.depth_0;
  public shadow_elevation_1 = shadows.depth_1;
  public shadow_elevation_2 = shadows.depth_2;

}
```

.. finally in you could then update the `ux-button-theme.css` to make use of these values:

styles.button.fab {
  border-radius: 50%;
  overflow: hidden;
  transition: box-shadow 0.2s cubic-bezier(0.4, 0, 0.2, 1);
  transition-delay: 0.2s;
  box-shadow: ${$design.shadow_elevation_2};
}

EisenbergEffect:
Bingo! Sounds great. Would you mind submitting a PR?

moodledan:
Sure, I'll mock it out over the next couple of days.

moodledan:
See PR here: https://github.com/aurelia/ux/pull/26

EisenbergEffect:
Thanks! We'll take a look shortly.

khuongduybui:
[...]

PR #53, opened 2017-04-08
feat(button): icon type

ZHollingshead:
addition to the button class enabling icon buttons to be made that are good for displaying in a list as a secondary action or in toolbars where buttons do not have backgrounds and need to be square

Usage:
``` html
<ux-button type="icon">
  <i class="material-icons more_vert"></i>
</ux-button>
```

resolves: #25

### DIFF
--- a/src/button/ux-button-theme.css
+++ b/src/button/ux-button-theme.css
@@ -76,6 +76,18 @@ styles.button.flat:active {
   background-color: ${background || $swatches.grey.p300};
 }
 
+styles.button.icon {
+  display: inline-flex;
+  align-content: center;
+  justify-content: center;
+
+  width: 40px;
+  height: 40px;
+
+  background-color: transparent;
+  color: ${foreground || $design.accent};
+}
+
 styles.button.fab {
   border-radius: 50%;
   overflow: hidden;
--- a/src/button/ux-button-theme.css
+++ b/src/button/ux-button-theme.css
@@ -85,7 +85,11 @@ styles.button.icon {
   height: 40px;
 
   background-color: transparent;
-  color: ${foreground || $design.accent};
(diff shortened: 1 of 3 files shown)


ZHollingshead:
The `<i class="material-icons more_vert"></i>` is just an example of an icon. Swap it out with Font Awesome or your choice of font.

shannonhorn:
got it, thx!

src/button/ux-button-theme.css
```diff
@@ -76,6 +76,22 @@ styles.button.flat:active {
   background-color: ${background || $swatches.grey.p300};
 }
 
+styles.button.icon {
```
EisenbergEffect: How does the icon class get onto the inner button?

src/button/ux-button-theme.css
```diff
@@ -76,6 +76,22 @@ styles.button.flat:active {
...
+  display: inline-flex;
+  align-content: center;
+  justify-content: center;
+
+  width: 40px;
+  height: 40px;
+
+  background-color: transparent;
+  color: ${foreground || $swatches.grey.p700};
```
EisenbergEffect: How does the icon get "into" the button?

ZHollingshead: See the usage above. Currently since we do not have our own icons it just centers an icon element the user can include. Down the line, I would prefer to have a usage similar to.

``` html
<ux-button type="icon" icon="icon-here"></ux-button>
```

EisenbergEffect: This usage scenario would be nice. We can circle back once we get the broader icon story resolved.

ZHollingshead: Alright, I will close this PR for the time being.

PR #56, opened 2017-04-14
feat(checkbox): add ux-checkbox control

ZHollingshead:
Adds the ux-checkbox control with the following features.

- Disabled state
- Binding to bool
- Binding to Array of strings
- Binding to Array of Objects
- Binding to Array with a Matcher
- Tab to select, enter to check
- Focused state
- Disabled state
- Material Design (with ripple)


jdanyow:
looks awesome- should `value`'s binding mode be one way?

ZHollingshead:
The result binds to the checked value so I do not believe so. I think the only time value might change is in a repeat.for.

src/checkbox/ux-checkbox.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (Array.isArray(this.checked)) {
+      isChecked = this.checked.some(item => this.matcher(item, elementValue));
+    }
+
+    if (isChecked) {
+      this.element.classList.add('checked');
+      this.element.setAttribute('aria-checked', 'true');
+    } else {
+      this.element.classList.remove('checked');
+      this.element.setAttribute('aria-checked', 'false');
```
EisenbergEffect: It makes me happy to see you thinking about accessibility 😄

PR #76, opened 2017-06-05
fix(button): add disable button behavior

### LINKED ISSUE #75: ux-button click event still fires despite being disabled

* **Library Version:**
0.2.0

* **Operating System:**
Windows 10

* **Node Version:**
7.4.0

* **NPM Version:**
4.3.0

Neither, using Aurelia CLI

* **Browser:**
all

* **Language:**
TypeScript 2.3.4

**Current behavior:**

If the user adds the "disabled" attribute to the <ux-input></ux-input> component, the css represents that it's disabled, but the click event still occurs.

The button's click event does not fire. So the "disabled" behavior includes both css change and behavior change in the button, just like the normal <button></button> tag

To Reproduce, simply use the following ux-button in your html and test your app.

```html
<ux-button disabled="true" click.delegate="delete()">Disabled Button</ux-button>
```

And add the delete function in your ts file.  In that function, just have a simple console.log message to demonstrate that the click handler was processed while the button was disabled.

The button's click event does not fire. So the "disabled" behavior includes both css change and behavior change in the button, just like the normal <button></button> tag

To have the ux-button component behave in the same fashion as the `<button></button>` tag.

chrsmrtn-:
this adds in disabled behavior when defined for `<ux-button></ux-button>`

resolves #75

### DIFF
--- a/src/button/ux-button.ts
+++ b/src/button/ux-button.ts
@@ -12,7 +12,7 @@ export class UxButton implements Themable {
   @bindable public type = null;
   @bindable public size = null;
   @bindable public effect = null;
-  @bindable public disabled: any = false;
+  @bindable public disabled: boolean | string = false;
   @bindable public theme = null;
 
   public view: View;
@@ -30,18 +30,29 @@ export class UxButton implements Themable {
       this.styleEngine.applyTheme(this, this.theme);
     }
 
-    if (this.disabled || this.disabled === '') {
+    // ensure we cast empty string as true
+    if (typeof this.disabled === 'string' && this.disabled === '') {
+      this.disabled = true;
+    }
+
+    if (this.disabled) {
       this.button.setAttribute('disabled', '');
     }
+
   }
 
   public themeChanged(newValue: any) {
     this.styleEngine.applyTheme(this, newValue);
   }
 
-  public disabledChanged(newValue: any) {
-    if (newValue === true || newValue === '') {
-      this.button.setAttribute('disabled', 'true');
+  public disabledChanged(newValue: boolean | string) {
+    // ensure we cast empty string as true
+    if (typeof this.disabled === 'string' && this.disabled === '') {
+      newValue = true;
+    }
+
+    if (newValue) {
+      this.button.setAttribute('disabled', '');
     } else {
       this.button.removeAttribute('disabled');
     }
--- a/src/button/ux-button.ts
+++ b/src/button/ux-button.ts
@@ -47,7 +47,7 @@ export class UxButton implements Themable {
 
   public disabledChanged(newValue: boolean | string) {
     // ensure we cast empty string as true
-    if (typeof this.disabled === 'string' && this.disabled === '') {
+    if (typeof newValue === 'string' && newValue === '') {
       newValue = true;
     }
 
--- a/src/button/ux-button.ts
+++ b/src/button/ux-button.ts
@@ -12,7 +12,7 @@ export class UxButton implements Themable {
   @bindable public type = null;
   @bindable public size = null;
   @bindable public effect = null;
-  @bindable public disabled = false;
+  @bindable public disabled: any = false;
   @bindable public theme = null;
 
   public view: View;
@@ -29,10 +29,22 @@ export class UxButton implements Themable {
     if (this.theme) {
       this.styleEngine.applyTheme(this, this.theme);
     }
+
+    if (this.disabled || this.disabled === '') {
+      this.button.setAttribute('disabled', '');
+    }
   }
 
   public themeChanged(newValue: any) {
     this.styleEngine.applyTheme(this, newValue);
+  }
+
+  public disabledChanged(newValue: any) {
+    if (newValue === true || newValue === '') {
+      this.button.setAttribute('disabled', 'true');
+    } else {
+      this.button.removeAttribute('disabled');
+    }
   }
 
   public onMouseDown(e: MouseEvent) {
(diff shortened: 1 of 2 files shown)


src/button/ux-button.ts
```diff
@@ -12,7 +12,7 @@ export class UxButton implements Themable {
   @bindable public type = null;
   @bindable public size = null;
   @bindable public effect = null;
-  @bindable public disabled = false;
+  @bindable public disabled: any = false;
```
RichiCoder1: Couldn't this be `  @bindable public disabled: boolean | string = false`?

src/button/ux-button.ts
```diff
@@ -29,12 +29,24 @@ export class UxButton implements Themable {
     if (this.theme) {
       this.styleEngine.applyTheme(this, this.theme);
     }
+
+    if (this.disabled || this.disabled === '') {
```
RichiCoder1: I'm curious what the reasoning is behind this line? It seems contradictory.

src/button/ux-button.ts
EisenbergEffect: @ZHollingshead Let's definitely use the typings mentinoed by @RichiCoder1

src/button/ux-button.ts
EisenbergEffect: I think it's to treat empty strings as true? It is confusing. @ZHollingshead If we switch to the more explicit typing above we can probably write some clearer logic here as well.

src/button/ux-button.ts
ZHollingshead: I like this a lot.

ZHollingshead: Let's go ahead and change this as mentioned above.

src/button/ux-button.ts
ZHollingshead: We should be able to check directly against false or empty with stricter types.

chrsmrtn-: @ZHollingshead, hey wanted to run this possible change past ya and get feedback.  I've casted the disabled variable per @RichiCoder1's suggestion:

```
   @bindable public disabled: boolean | string = false;
```

Now I'm working on the subsequent if block to clear that logic up a bit.  Specifically, I wanted to highlight that we are attempting to detect an empty string as true.  All other casting from string to boolean will behave as normal. As well as all other castings to boolean will behave as normal.

```
    // ensure we cast empty string as true
    if (typeof this.disabled === 'string' && this.disabled === '') {
      this.disabled = true;
    }

    if (this.disabled) {
      this.button.setAttribute('disabled', '');
    }
```

Also wondering if we should be using `this.button.setAttribute('disabled', 'true');` instead like we've done in the `disabledChanged` hander?

src/button/ux-button.ts
```diff
@@ -29,12 +29,24 @@ export class UxButton implements Themable {
...
+      this.button.setAttribute('disabled', '');
+    }
   }
 
   public themeChanged(newValue: any) {
     this.styleEngine.applyTheme(this, newValue);
   }
 
+  public disabledChanged(newValue: any) {
```
chrsmrtn-: Should we also adjust this casting to match what we've done for the disabled attribute?  So this would now become:
`public disabledChanged(newValue: boolean | string)`?

src/button/ux-button.ts
RichiCoder1: That logic makes much more sense now. Is there a meaningful difference in browser between `disabled=""` and `disabled="true"`? If not, I'd opt for the former in both places.

chrsmrtn-: From what's I've found. It seems that disabled="true" will result to false.  This stackoverflow answer references the HTML 5 spec:

https://stackoverflow.com/questions/6961526/correct-value-for-disabled-attribute

I'm almost thinking now that the original intention of that snip of code was to follow the HTML 5 spec.  I took my original code from the ux-input.ts source to stay consistent within the project.

@EisenbergEffect @ZHollingshead, does the above sound right when ux-input was written out?

After we finalize this behavior and coding for the ux-button, i'll make another pull request to move these changes into other components that use the disabled and readonly attributes.

src/button/ux-button.ts
RichiCoder1: Yup! Though in this case it'd be `boolean | string | null` as I believe an end user could technically null it out.

src/button/ux-button.ts
RichiCoder1: Ahh, yes, that's the correct SO I was looking for. Boolean attributes in HTML are a bit weird about that. I'd def do `setAttibute(`attributeName`, '')`  or all instances of disabled and readonly. I don't think Aurelia is too concerned with XHTML compatibility.

ZHollingshead: @chrsmrtn- `setAttribute('disabled', '')` is the way to go in my opinion.

src/button/ux-button.ts
chrsmrtn-: This appears to be called by the binding stuff within Aurelia.  I'm going to move this to `boolean | string` for now and align the if statement to match for clarity. If something else comes later on down the road we can re-visit this bit.

src/button/ux-button.ts
```diff
@@ -29,12 +29,35 @@ export class UxButton implements Themable {
...
+
   }
 
   public themeChanged(newValue: any) {
     this.styleEngine.applyTheme(this, newValue);
   }
 
+  public disabledChanged(newValue: boolean | string) {
+    // ensure we cast empty string as true
+    if (typeof this.disabled === 'string' && this.disabled === '') {
```
RichiCoder1: Did you mean `if (typeof newValue === 'string' && newValue = '') {`?

chrsmrtn-: Yep that I did!

will fix to:

```
if (typeof newValue === 'string' && newValue === '') {
```

PR #80, opened 2017-06-09
fix(checkbox) disabled checkbox behavior

### LINKED ISSUE #79: ux-checkbox disabled behavior issues

* **Library Version:**
0.2.0

* **Operating System:**
Windows 10

* **Node Version:**
7.4.0

* **NPM Version:**
4.3.0

Neither, using Aurelia CLI

* **Browser:**
all

* **Language:**
TypeScript 2.3.4

**Current behavior:**

If the user adds the "disabled" attribute to the <ux-checkbox></ux-checkbox> component, the css represents that it's enabled, and the click event still occurs.

The checkbox's click event does not fire and the checkbox renders as "disabled". So the "disabled" behavior includes both css change and behavior change in the checkbox, just like the normal <input type="checkbox"></input> tag

To Reproduce, simply use the following ux-checkbox in your html and test your app.

```html
<ux-checkbox label="Check Option" disabled click.delegate="checkChange()"></ux-checkbox>
```

And add the `checkChange` function in your ts file.  In that function, just have a simple console.log message to demonstrate that the click handler was processed while the checkbox was disabled.

The checkbox's click event does not fire and the checkbox renders as "disabled". So the "disabled" behavior includes both css change and behavior change in the checkbox, just like the normal `<input type="checkbox"></input>` tag

To have the ux-checkbox component behave in the same fashion as the `<input type="checkbox"></input>` tag.

chrsmrtn-:
this fixes disabled behavior when defined for `<ux-checkbox></ux-checkbox>`

resolves #79

### DIFF
--- a/src/checkbox/ux-checkbox.ts
+++ b/src/checkbox/ux-checkbox.ts
@@ -32,7 +32,13 @@ export class UxCheckbox implements Themable {
   private ripple: PaperRipple | null = null;
 
   public get isDisabled() {
-    return this.disabled === true || this.disabled === '' || this.disabled === 'disabled';
+    let ret: boolean = this.disabled;
+    if (typeof this.disabled === 'string' &&
+        (this.disabled === '' || this.disabled.toString().toLocaleLowerCase() === 'disabled')) {
+      ret = true;
+    }
+
+    return ret;
   }
 
   constructor(public element: HTMLElement, public resources: ViewResources, private styleEngine: StyleEngine) { }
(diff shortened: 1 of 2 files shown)


src/checkbox/ux-checkbox.ts
```diff
@@ -32,7 +32,13 @@ export class UxCheckbox implements Themable {
   private ripple: PaperRipple | null = null;
 
   public get isDisabled() {
```
EisenbergEffect: If we are going to bind to this property, we should add a `@computedFrom('disabled')` decorator to this property.

EisenbergEffect: It might be a good idea anyway.

chrsmrtn-: Yeah, I'll get that in there. Thanks!

src/checkbox/ux-checkbox.ts
```diff
@@ -1,4 +1,5 @@
 import { customElement, bindable, ViewResources, View, processAttributes } from 'aurelia-templating';
+import { computedFrom } from 'aurelia-framework';
```
ZHollingshead: This is really a minor thing, but we should put computedFrom into the 'aurelia-binding' import rather than making a new one referencing 'aurelia-framework' to be consistent.

chrsmrtn-: That makes sense. I didn't know you can import that from aurelia-binding.  Thanks, will get this adjusted shortly.

Issue #83, opened 2017-06-29
ux-tabs


manuel-guilbault:
Implement the ux-tabs and ux-tab components for Material and iOS design.

## General Todo

For Material design:
- [x] basic style
- [ ] transition animation
- [ ] automatic sizing
- [ ] text wrap and ellipsis
- [ ] overflow pagination
- [ ] support for drop down menus

For iOS design:
- [ ] basic style

manuel-guilbault:
I already have a working version. It looks like this:

```html
<ux-tabs>
  <ux-tab text="Account" icon="face"></ux-tab>
  <ux-tab text="Favorites" icon="favorite"></ux-tab>
</ux-tabs>
```

ZHollingshead:
Issue moved to [aurelia/ux-components #6](https://github.com/aurelia/ux-components/issues/6) via [**ZenHub**](https://www.zenhub.com/)

xenoterracide:
this was reopened? so there are no tabs :'(

bigopon:
nagging @ben-girardet 😏

ben-girardet:
Great question @xenoterracide 

I joined this project recently and was also intrigued by what happened to the `tabs` component. I figured out from [this issue](https://github.com/aurelia/ux/pull/98) that it was removed back in Oct 17 from the library due to some bugs.

What need to be done ? The library has much evolved since then, but the original work on tabs might still be of interest. Wondering if @serifine could give us some insights ?

The course of action could be to 

1. trying to bring back the code of the tabs into the library today and evaluate if we can keep going from here ([here is the latest code I've found from this repo](https://github.com/aurelia/ux/tree/e1fa8c677585ec66c18077ad0eb52b0e68fb2e3c/src/tabs))
2. continue from there or start again from scratch ?!?

Also, the fork of the original author @manuel-guilbault [still exists here with demo of the tabs](https://github.com/manuel-guilbault/aurelia-app-ux-showcase/tree/ux-tabs).

@xenoterracide did you ask by curiosity or would you be willing to help out with this component ? If yes, let's talk about how we could include you in the process

xenoterracide:
curiousity, and potential (but not promised) willingness to help out. I'm still fairly green again at doing Web UI stuff. Currently was debating on whether I wanted to try Aurelia UX in addition to my current Bulma experiment (trying to avoid frameworks with that jquery thing).

ben-girardet:
[...]

PR #84, opened 2017-06-29
feat(ux-tabs): create ux-tabs and ux-tab elements

manuel-guilbault:
linked to #83


manuel-guilbault:
I also added some examples in the ux showcase app: https://github.com/manuel-guilbault/aurelia-app-ux-showcase

src/tabs/ux-tab-theme.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+styles.tab:hover:not(.disabled) {
+  opacity: 1;
+}
+
+styles.tab.selected:not(.disabled) {
+  opacity: 1;
+  padding-bottom: 18px;
+  border-bottom: ${foreground || $design.accentForeground} solid 2px;
```
ZHollingshead: The indicator should be `$design.accent` if we are going to base this on the accent color. `$design.accentForeground` is not always `#fff`. In addition, I would also like to see `foreground` changed to a new property specific to the indicator that can be set in the .ts theme.

> Indicator
>
> - Height: 2dp
> - Color: #fff or accent color

[Material Guidelines - Tabs](https://material.io/guidelines/components/tabs.html#tabs-specs)

src/tabs/ux-tab-theme.css
```diff
@@ -0,0 +1,34 @@
+styles.tab {
+  display: inline-block;
+  text-transform: uppercase;
+  cursor: pointer;
+  opacity: 0.7;
+  padding: 14px 12px 20px 12px;
+  text-align: center;
+  color: ${foreground || $design.accentForeground};
+  background-color: ${background || $design.accent};
```
ZHollingshead: I would not base the background on the accent.

Try leaving this as either transparent or set it to the primary color.

manuel-guilbault:
@ZHollingshead so with icons, it would give something like this:

```html
<ux-tabs>
  <ux-tab><ux-icon icon="face"></ux-icon> Account</ux-tab>
  <ux-tab><ux-icon icon="favorite"></ux-icon>Favorites</ux-tab>
</ux-tabs>
```

Is this what you have in mind?

ZHollingshead:
I was thinking just for text content. I like how the icon is rendered it feels clean. This is more as an optional way to add text or content to a tab. So one could either do `<ux-tab>Text</ux-tab>` or `<ux-tab text="Text"></ux-tab>`. I hope that makes sense.

This should be achievable by adding a slot. If the below snippet does not work then just move the binding outside of the slot, I cannot remember off the top of my head if you can use a binding as default slot content.

In src/tabs/ux-tab.html
``` html
<span if.bind="text" class="text"><slot>${text}</slot></span>
```

So I am thinking good usage could be:
``` html
<ux-tabs>
  <ux-tab>Account</ux-tab>
  <ux-tab>Favorites</ux-tab>
</ux-tabs>

<ux-tabs>
  <ux-tab icon="face"> Account</ux-tab>
  <ux-tab icon="favorite">Favorites</ux-tab>
</ux-tabs>

<ux-tabs>
  <ux-tab icon="face" text="Account"></ux-tab>
  <ux-tab icon="favorite" text="Favorites"></ux-tab>
</ux-tabs>
```

manuel-guilbault:
Ok makes total sense. I'll update my pull request in a couple of minutes.

On another topic, I would like to animate the indicator when the selected tab changes, but I'll need some help with that, as I don't know much about low-level animations (I always used libraries like jQuery or Velocity). Any idea on how this could be implemented? Any resources which could point me in the right direction?

ZHollingshead:
For animation take a look at these. They are not bad for getting started with animation. One thing to keep in mind is always using `transform` based animations, as those take advantage of hardware acceleration.

https://css-tricks.com/almanac/properties/a/animation/
https://css-tricks.com/almanac/properties/t/transform/

manuel-guilbault:
I just pushed some updates. Let me know if there's anything I forgot or misunderstood.

Thanks for the links. I'll give it a look and let you know how it goes :)

manuel-guilbault:
@ZHollingshead I just pushed a first attempt at animating the indicator. It works well, except for a weird glitch: when the showcase app loads, there's a small duration (a fraction of a second) during which the whole app seems to be rendered bigger, then every element on the page shrinks in size. Because of this, the indicator is initially rendered too big. I don't understand what is the cause of this. I'll keep digging on my end, but if you have any idea of why this happens, let me know ;)

PR #88, opened 2017-08-02
refactor(components): disabled and readonly properties

chrsmrtn-:
refactored the readonly and disabled property logic based on feedback in pull request #76.

- clearer intent of the code when converting string values to Boolean for Boolean attributes
- now handles string values of 'readonly' and 'disabled' to represent a true value (http://www.w3.org/TR/html5/infrastructure.html#boolean-attributes)

### DIFF
--- a/src/resources/html-attributes.ts
+++ b/src/resources/html-attributes.ts
@@ -1,0 +1,17 @@
+// this function will ensure that we propertly treat a potential string value for a boolean attribute
+// as the boolean representation
+export function normalizeBooleanAttribute(attributeName: string, value: boolean | string): boolean {
+  let ret: boolean;
+  if (typeof value === 'string') {
+    if (value === '' || value.toLocaleLowerCase() === attributeName.toLocaleLowerCase()) {
+      // if string, then it can be true if the value is blank,
+      // or the value matches the name of the attribue with case insensitivity
+      ret = true;
+    } else {
+      ret = false;
+    }
+  } else {
+    ret = value as boolean;
+  }
+  return ret;
+}
--- a/src/resources/html-attributes.ts
+++ b/src/resources/html-attributes.ts
@@ -1,5 +1,12 @@
-// this function will ensure that we propertly treat a potential string value for a boolean attribute
-// as the boolean representation
+/**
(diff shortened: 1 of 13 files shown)


chrsmrtn-:
That's a great idea. That logic for translating a boolean html attribute will be same for any boolean html attribute. So in the future, if we need to work with other boolean html attributes, then we can leverage one bit of code to make that translation.

I'll take a look and see where this "helper function" may fit. If you have any ideas off hand based on your previous work, let me know.

chrsmrtn-:
Hey, resolved a conflict, but still have to look into refactoring the boolean html attribute conversion logic

chrsmrtn-:
Okay centralized the html Boolean attribute value check logic.  Also instead of changing `this.disabled` or `this.readonly` property values, we now only normalize the value and make our decision based on the resulting true or false. So we leave the original property value alone, but treat it as we should in the code. That way users of our components will not see their 'disabled' value suddenly change to a Boolean true after the component has been loaded.

If it's preferred that we continue to modify the property values, I can change it back. I wanted to make all of the changes clear with this pull request update.

src/resources/html-attributes.ts
```diff
@@ -0,0 +1,17 @@
+// this function will ensure that we propertly treat a potential string value for a boolean attribute
+// as the boolean representation
+export function normalizeBooleanAttribute(attributeName: string, value: boolean | string): boolean {
```
ZHollingshead: This may feel nitpicky, but can we change the comment to a jsdoc style comment. A small change, but this would give us IntelliSense support in most editors.

``` javascript
/*
 * This function will ensure that we properly treat a potential string
 * value for a boolean attribute as the boolean representation
 */
```

ZHollingshead: I just feel if we are going to have the comment there we should make the most of it.

chrsmrtn-: Sweet! Yeah! I didn't know that a comment in that style at that location would also be picked up by various editors.

chrsmrtn-:
Okay changes are in. I wanted to reword my commit message as I forgot to include doc(<scope>) in my comment.  But it didn't seem to take.

PR #98, opened 2017-10-06
Aurelia UX Overhaul

### LINKED ISSUE #50: [suggestion] dont use .css extension for non-valid dynamic css files
today i tried aurelia-ux with the aurelia-webpack-skeleton, this currently doesnt work, i wasnt able to dig deeper into the issues, but one thing seemed to be clear: dynamic css files are not valid css, so they cannot be processed by css loaders; second: the common way to configure loaders is to pass the file extension, which is .css. i know there are ways to exclude files, but with the current webpack skeleton and the config-helper easy-webpack/config-css this isnt really possible right now.

...how to bundle dynamic css files with webpack is then a different issue...

### LINKED ISSUE #81: Use Aurelia-UX(theme) within skeleton-esnext-webpack got a error
Hello friends, i using the newest skeleton-esnext-webpack and  aurelia-ux, but got a error when build:
```
ERROR in ./src/app.html
Module not found: Error: Can't resolve 'common/styles/common-ux.css#ux' in '/Users/xujin/My_Projects/NodeJS/skeleton-esnext-webpack/src'
 @ ./src/app.html
 @ ./src/app.js
 @ ./src/main.js
 @ ./~/aurelia-webpack-plugin/dist/aurelia-entry.js
 @ multi aurelia-webpack-plugin/dist/aurelia-entry aurelia-polyfills aurelia-loader-webpack (webpack)-dev-server/client?http://localhost:8080 aurelia-bootstrapper
```
I try to rename `common-ux.css` to `common-ux.css#ux`, but got another error:
```
ERROR in ./src/common/styles/common-ux.css#ux
Module build failed: Unknown word (16:23)

  14 | }
  15 | styles.primary {
> 16 |   background-color: ${$design.primary};
     |                       ^
  17 | }
  18 | styles.left_font_color {
  19 |   color: ${$design.left_font_color};

 @ ./src/app.html
 @ ./src/app.js
 @ ./src/main.js
 @ ./~/aurelia-webpack-plugin/dist/aurelia-entry.js
 @ multi aurelia-webpack-plugin/dist/aurelia-entry aurelia-polyfills aurelia-loader-webpack (webpack)-dev-server/client?http://localhost:8080 aurelia-bootstrapper
``` 

main.js fragment
```javascript
import 'babel-polyfill';
import * as Bluebird from 'bluebird';
import i18nextBackend from 'i18next-xhr-backend';
import moment from 'moment';
import {AppRouter} from 'aurelia-router';

Bluebird.config({ warnings: { wForgottenReturn: false } });

export async function configure(aurelia) {
    aurelia.use
        .standardConfiguration()
        .developmentLogging()
        .plugin(PLATFORM.moduleName('aurelia-animator-css'))
        .plugin(PLATFORM.moduleName('aurelia-ux'))
        .feature(PLATFORM.moduleName('common'));
        ......
```

Node: 6.11.0
aurelia-ux: 0.2.0
skeleton-esnext-webpack: the master branch

### LINKED ISSUE #97: Use CSS Variable style binding
Change binding syntax in CSS files to bind to css variables.

**What is the motivation / use case for changing the behavior?**
This will make our bindings parse as valid CSS, solving issues we currently have with Webpack and PostCSS exhibited in #50 and #81.

**Requirements**

> *W3C CSS Variable Structure*
> ```
> var( <custom-property-name> [, <declaration-value> ]? )
> ```

Aurelia UX should parse the incoming stylesheet and set up and replace CSS variables at runtime.

In the future parsing of the stylesheet will become unnecessary when CSS Variables become more widely supported so the implementation in the stylesheets should be compliant with current CSS Variables

**Example**
```css
test {
  margin: 0;
  height: 100%;
  background-color: ${ foreground || '#fff'};
  color: #fff;
  flex: 1;
}
```
Becomes
```css
test {
  margin: 0;
  height: 100%;
  background-color: var(--foreground, #fff);
  color: #fff;
  flex: 1;
}
```

### LINKED ISSUE #99: Enable individual installation of components
@ZHollingshead commented on [Sat Sep 02 2017](https://github.com/aurelia/ux-components/issues/21)

As part of the move to separate the components into a separate library, each component should now include all the required pieces to make a self-packaged component that can be installed individually in the use case where a user may not want to install the entire component suite.

The following will need to be accomplished for each component:
1. Create a `package.json` file containing the build script and package information
2. Create an `index.ts` file that is the entry point of the component when used individually

### Requirements
Aurelia UX Components should be able to be installed in one of two ways.

Option one, install the entire suite.
`npm install aurelia-ux-components --save`

Option two, install components individually.
`npm install aurelia-ux-components/component-name`

### Progress Tracker
This section will be updated as tasks are completed
#### Components
- [ ] Button
- [ ] Checkbox
- [ ] Chip Input
- [ ] Datepicker
- [ ] Form
- [ ] Icons
- [ ] Input
- [ ] Input Info
- [ ] Textarea
- [ ] List
- [ ] Tabs

---

@ZHollingshead commented on [Mon Sep 11 2017](https://github.com/aurelia/ux-components/issues/21#issuecomment-328552194)

Considering leveraging Lerna for managing the monorepo.

Still trying to figure out how to best build the split components.

ZHollingshead:
### Build Process

These commands will need to be run in order

`npm run build` at the base project to build Aurelia UX
`lerna bootstrap` sets up a symlink between all of the packages in the monorepo
`lerna run build` builds all of the mono repo projects. 

> Note: `lerna run build` is very CPU intensive and takes a small period of time on most machines. If you are working within a single component, you might try `npm run build` instead at the component level.

### Core Changes
* Style Engine has been completely rewritten
* The library now uses CSS Variables natively
* CSS Modules are not in place, as themes are now natively scoped using CSS Variables now. This can be added later if it is viewed as a good feature, but it needs to be rewritten if we are to add it and the scope of this change is already very large.

### Components
* All components moved back into the Aurelia-UX repository using the Monorepo design
* Tab component has been removed. This control was buggy, did not style right in the end and just while being used does not match the quality of the other controls.
* UX Icon component is still here, however,  I have started to directly use SVG's in the components for items such as button icons in other controls to prevent dependencies.

### Related Issues
Resolves #99 
Resolves #97 
Resolves #81 
Resolves #50 
Resolves #99


ZHollingshead:
@EisenbergEffect I added a couple things to the style engine to make working with the default themes easier.

Style engine usage

```typescript
  // retrieves the bound instance of the default theme
  let theme = this.styleEngine.getDefaultTheme('button');

  // changes here will take effect immediately
  theme.foreground = "#363636";
  theme.background = "#fff";
```

```typescript

  const theme = new UxListTheme();

  theme.listBackground = '#232323';
  theme.listForeground = '#656565';
  theme.listSecondaryForeground = '#565656';

  // will ensure that the theme has been created before applying the theme, useful at application startup before any controls have been loaded
  this.styleEngine.applyTheme(theme);
```

**Theme Groups**

Theme groups were created as an easy way to make reusable themes. A theme can be created and passed into the applyThemes method on the style controller.

```typescript
class UxThemeExample {
  public static getTheme() {
    let themes: UxTheme[] = [];

    const listTheme = new UxListTheme();

    listTheme.listBackground = '#232323';
    listTheme.listForeground = '#656565';

    themes.push(listTheme);

    const buttonTheme = new UxButtonTheme();

    buttonTheme.background = '#232323';
    buttonTheme.foreground = '#656565';

    themes.push(buttonTheme);

    return themes;
  }
}

// some view

  public setNewTheme() {
    this.styleEngine.applyThemeGroup(UxThemeExample.getTheme());
  }
```

EisenbergEffect:
I'm totally swamped during the week these days. I'll review this weekend. Love the theme API discussed above. Looking forward to this. Very cool stuff!

packages/chip-input/src/ux-chip.css
```diff
@@ -0,0 +1,48 @@
...
+    margin-right: 0;
+  }
+
+    ux-chip[deletable] > span.close {
+      display: inline-flex;
+      justify-content: center;
+      align-items: center;
+      margin: 0 4px;
+      color: var(--ux-theme--chip-delete-foreground);
+      background-color: var(--ux-theme--chip-delete-background);
```
EisenbergEffect: I'm wondering about the variable naming convention. Why two dashed between ux-theme and chip but above only one dash between design and elevation4dp?

packages/datepicker/package.json
```diff
@@ -0,0 +1,133 @@
+{
+  "name": "@aurelia-ux-components/datepicker",
+  "version": "0.0.1",
+  "description": "Checkbox component for Aurelia UX",
```
EisenbergEffect: Make sure to check all package metadata. I just noticed this one isn't quite right.

packages/datepicker/src/ux-calendar.css
```diff
@@ -0,0 +1,55 @@
...
+
+ux-calendar div.calendar-row {
+  display: flex;
+  align-items: center;
+  justify-content: space-around;
+  height: 2em;
+}
+
+ux-calendar .calendar-row.weekdays {
+  color: var(--ux-theme--datepicker-weekday-foreground, var(--swatch-grey-p600));
```
EisenbergEffect: What does the nested var expression do?

packages/form/src/ux-field.ts
```diff
@@ -0,0 +1,21 @@
...
+export class UxField {
+  @bindable public label: string;
+
+  constructor(private element: Element) { }
+
+  public attached() {
+    if (this.label && !this.element.querySelector('label')) {
+      const newLabel = document.createElement('label');
+
+      newLabel.textContent = this.label;
```
EisenbergEffect: If someone databinds the label property, the generated/located label will need to update in response to that.

packages/button/src/ux-button.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  @bindable public theme: UxButtonTheme;
+
+  private ripple: PaperRipple | null = null;
+  private button: HTMLButtonElement;
+
+  constructor(
+    public element: HTMLElement,
+    private styleEngine: StyleEngine) {
+      styleEngine.ensureDefaultTheme(new UxButtonTheme());
```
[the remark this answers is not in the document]
ZHollingshead: In the constructor of the theme?

packages/button/src/ux-button.css
```diff
@@ -0,0 +1,126 @@
...
+
+ux-button > button.raised.large, ux-button > button.flat.large, ux-button > button.icon.large {
+  font-size: 16px;
+  height: 42px;
+  line-height: 42px;
+}
+
+ux-button > button.raised, ux-button > button.fab {
+  background-color: var(--theme-background, var(--design-accent));
+  color: var(--button-foreground, var(--design-accent-foreground));
```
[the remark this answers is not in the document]
ZHollingshead: Needs to be --ux-theme--button, I missed that.

packages/datepicker/src/ux-calendar.css
ZHollingshead: With css variables, to make a fallback that is also a variable, it has to be wrapped in a var().

var(/*variable name*/, /*fallback property*/) 

So if a fallback property is also a css variable, it must also be wrapped.

packages/chip-input/src/ux-chip.css
ZHollingshead: This was update to `--ux-design--` as the prefix to match up with the `--ux-theme--` prefix.

ZHollingshead:
@EisenbergEffect just finished testing this, ready for a second pass.

packages/icons/package.json
```diff
@@ -0,0 +1,131 @@
...
+  "license": "MIT",
+  "author": "Rob Eisenberg <rob@bluespire.com> (http://robeisenberg.com/)",
+  "contributors": [
+    "Zach Hollingshead <zach.hollingshead@gmail.com>"
+  ],
+  "main": "dist/commonjs/index.js",
+  "typings": "dist/commonjs/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aurelia/ux-components"
```
EisenbergEffect: Also, noticed that this url needs to be fixed.

packages/textarea/src/html-attributes.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @description This function will ensure that we propertly treat a potential string value for a boolean attribute
+ * as the boolean representation
+ *
+ * @param {string} attributeName Name of the boolean attribute we are normalizing for
+ * @param {boolean|string} value Existing value of the boolean html attribute represented as a boolean or string
+ *
+ * @returns {boolean}
+ */
+export function normalizeBooleanAttribute(attributeName: string, value: boolean | string): boolean {
```
EisenbergEffect: Is this a duplicate?

src/styles/style-engine.ts
```diff
@@ -1,94 +1,61 @@
...
+   * and applies them to the provided element.
+   *
+   * @param element Element to apply the processed UxTheme to.
+   * @param theme UxTheme to process.
+   */
+  public applyTheme(theme: UxTheme, element?: HTMLElement) {
+    if (element != null) {
+      for (const key in this.styleController.getThemeKeys(theme)) {
+        if (theme[key] != null) {
+          element.style.setProperty(`this.generateCssVariable(theme.themeKey, key, theme[key])`, theme[key]);
```
EisenbergEffect: I think there's more to it than this. For example, if it's a new theme instance, then we need to observe the properties on the new theme object for changes so that we can update the local css variables based on changes to the local theme.

EisenbergEffect: It's even trickier than that because it's possible that it's a shared theme object, but not the default object. So, we don't need to subscribe the second time, but there's an additional set of local css variable to update on changes. Does that make sense?

packages/button/src/ux-button-theme.ts
```diff
@@ -0,0 +1,25 @@
...
+  public flatForeground: string = 'var(--ux-design--primary)';
+
+  public accentBackground: string = 'var(--ux-design--accent)';
+  public accentForeground: string = 'var(--ux-design--accent-foreground)';
+
+  public accentFlatBackground: string = 'transparent';
+  public accentFlatForeground: string = 'var(--ux-design--accent)';
+
+  public backgroundDisabled: string = swatches.grey.p500;
+  public foregroundDisabled: string = swatches.grey.p100;
```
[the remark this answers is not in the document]
ZHollingshead: Fortunately, we already have them in CSS Variable form!

PR #121, opened 2017-12-19
feat(switch): Add switch component

ZHollingshead:
Add a switch component for Aurelia UX

image

## Usage
```html
<ux-switch checked.bind="booleanValue"></ux-switch>
```

### DIFF
--- a/packages/switch/src/ux-switch.ts
+++ b/packages/switch/src/ux-switch.ts
@@ -66,7 +66,7 @@ export class UxSwitch implements UxComponent {
 
   public themeChanged(newValue: UxSwitchTheme) {
     if (newValue != null && newValue.themeKey == null) {
-      newValue.themeKey = 'radio';
+      newValue.themeKey = 'switch';
     }
 
     this.styleEngine.applyTheme(newValue, this.element);
--- a/packages/switch/src/ux-switch.ts
+++ b/packages/switch/src/ux-switch.ts
@@ -1,0 +1,112 @@
+import { customElement, bindable, ViewResources } from 'aurelia-templating';
+import { computedFrom, bindingMode } from 'aurelia-binding';
+import { inject } from 'aurelia-dependency-injection';
+import { StyleEngine, UxComponent, normalizeBooleanAttribute } from '@aurelia-ux/core';
+import { UxSwitchTheme } from './ux-switch-theme';
+
+@inject(Element, ViewResources, StyleEngine)
+@customElement('ux-switch')
+
+export class UxSwitch implements UxComponent {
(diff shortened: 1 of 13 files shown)


packages/switch/src/ux-switch.ts
```diff
@@ -0,0 +1,112 @@
...
+  @bindable({ defaultBindingMode: bindingMode.twoWay })
+  @bindable public checked: boolean = false;
+
+  @computedFrom('disabled')
+  public get isDisabled() {
+    return normalizeBooleanAttribute('disabled', this.disabled);
+  }
+
+  constructor(public element: HTMLElement, public resources: ViewResources, private styleEngine: StyleEngine) {
+    styleEngine.ensureDefaultTheme(new UxSwitchTheme());
```
EisenbergEffect: Weren't we going to move this to an external shared theme instance?

packages/switch/src/ux-switch.ts
```diff
@@ -0,0 +1,112 @@
...
+        labelElement.removeEventListener('click', () => {
+          this.toggleSwitch();
+        });
+      }
+    }
+  }
+
+  public themeChanged(newValue: UxSwitchTheme) {
+    if (newValue != null && newValue.themeKey == null) {
+      newValue.themeKey = 'radio';
```
EisenbergEffect: What does this do?

ZHollingshead: This handles the case where the user is passing in a brand new theme. If they do not instantiate a new theme and just pass in an object, this could be lacking the themeKey. The theme key is used to build the CSS Variable names. So `radio` becomes. `--ux-theme--radio-etc`. However the theme key for this particular line is wrong. I need to change this.

packages/switch/src/ux-switch.ts
ZHollingshead: We looked at it but if I remember correctly, the CSS variables made that a moot point.

Right now when a component is created, UX will look for a registered default theme. If a default theme does not exist, this creates a new instance at the root with the appropriate CSS variables. If those CSS variables are changed, it immediately cascades to all controls in the document that use that CSS variable, aka the components that use the theme. This is unless the component is bound to a different theme, in which case it is using a set of CSS variables from that bound theme instead which overrides the default.

ZHollingshead:
@EisenbergEffect I pushed an update to resolve the incorrect key.

packages/switch/src/ux-switch.ts
EisenbergEffect: I thought we were using a shared instance just to avoid the unnecessary class instantiation, esp. in the case where there may be lots of components.

PR #130, opened 2017-12-21
Add css reset and global style controller

ZHollingshead:
Adds reset css code to the header of the document
when aurelia-ux is started. This can be disabled in the configuration.

combines all css variables into a single style element

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,5 +1,6 @@
 import { FrameworkConfiguration } from 'aurelia-framework';
 import { AureliaUX } from './aurelia-ux';
+import { PLATFORM } from 'aurelia-pal';
 
 export { swatches } from './colors/swatches';
 export { shadows } from './colors/shadows';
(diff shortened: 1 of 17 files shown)


packages/core/src/index.ts
```diff
@@ -12,16 +13,24 @@ export { UxComponent } from './components/ux-component';
...
 export { AureliaUX } from './aurelia-ux';
+export { UXConfiguration } from './ux-configuration';
 
 export function configure(config: FrameworkConfiguration, callback?: (config: AureliaUX) => Promise<any>) {
   const ux = config.container.get(AureliaUX) as AureliaUX;
 
+  config.globalResources([
+    PLATFORM.moduleName('./reset.css')
+  ]);
+
```
EisenbergEffect: I'm thinking that you want to use `PLATFORM.moduleName` for Webpack inspection here, but you don't actually want to pass it to the `globalResources` function, since that will cause it to be loaded no matter what. I think you really want that to be controlled by the configuration below.

ZHollingshead: So one question I have is this, without that line there the CSS file is not actually included with the code (at least in webpack). How can I ensure that the file is included when the app is built?

EisenbergEffect: I think you can just put `PLATFORM.moduleName('./reset.css')` in the code by itself and the webpack plugin will detect it and bundle.

ZHollingshead:
@EisenbergEffect Alright, I moved the `PLATFORM.moduleName(`'./reset.css');` inside of the configuration and removed it from the `globalResources`

ZHollingshead:
@EisenbergEffect is this good to merge?

EisenbergEffect:
Good to go!

PR #132, opened 2017-12-23
feat(UxSelect): ux-select element

bigopon:
* Multiple
* Single
* Keyboard navigation
* `blur`, `change`, `focus` events
* Ripple on list items
* Boolean attribute binding behavior (I think we need this in `@aurelia-ux/core`)
* disabled
* placeholder
* aria

* demo https://gist.run/?id=b18d8e26af15a315892a021d9c919d7e (awaiting feedback)

### todo

- [x] item disabled (need discussion about this)
- [ ] ? (help wanted for filling this list)

### hard to do:

* Option group (need some discussion here, totally unclear how we going to support this), without allowing `options` to be constructed by user.

### DIFF
--- a/packages/select/.npmignore
+++ b/packages/select/.npmignore
@@ -1,0 +1,2 @@
+src
+tsconfig.json
--- a/packages/select/package.json
+++ b/packages/select/package.json
@@ -1,0 +1,136 @@
+{
+  "name": "@aurelia-ux/select",
+  "version": "0.6.0",
+  "description": "Select component for Aurelia UX",
+  "keywords": [
+    "aurelia",
+    "styles",
+    "components",
+    "ux",
+    "selec",
+    "dropdown"
+  ],
+  "homepage": "http://aurelia.io",
+  "bugs": {
+    "url": "https://github.com/aurelia/ux/issues"
+  },
+  "license": "MIT",
+  "author": "Rob Eisenberg <rob@bluespire.com> (http://robeisenberg.com/)",
+  "contributors": [
+    "bigopon <bigopon@gmail.com>"
+  ],
+  "main": "dist/commonjs/index.js",
+  "typings": "dist/commonjs/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aurelia/ux"
+  },
+  "scripts": {
+    "lint": "cross-env tslint --project tsconfig.json",
+    "pretest": "cross-env npm run lint",
+    "test": "cross-env rimraf dist && tsc && karma start --single-run",
+    "develop": "concurrently \"./node_modules/.bin/tsc --watch\" \"./node_modules/.bin/karma start\"",
+    "prebuild:amd": "cross-env rimraf dist/amd",
+    "build:amd": "cross-env tsc --project tsconfig.json --outDir dist/amd --module amd",
+    "postbuild:amd": "cross-env copyfiles --up 1 src/**/*.html src/**/*.css dist/amd",
+    "prebuild:commonjs": "cross-env rimraf dist/commonjs",
+    "build:commonjs": "cross-env tsc --project tsconfig.json --outDir dist/commonjs --module commonjs",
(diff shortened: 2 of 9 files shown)


EisenbergEffect:
Awesome, solid start! @ZHollingshead to review please :)

ZHollingshead:
This already looks awesome! So as far as adding to the list, it would be nice to see a placeholder attribute for filling in the empty state. Also can we make the open animation a bit faster? I think Material animation durations are either 125ms or 250ms.

I also think the multiple should be a simiple list of options that you can select from, similar to the multiple select html control so we stay close to what a user would expect.. What do you guys think? 

Again, this is looking great.

bigopon:
> placeholder ...

It was a bit painful to update the gist, but now the demo has placeholder.

> 125/ 250

I didn't know that, fixed. It feels a lot cooler than 200s 👍 

> I also think the multiple should be a simple list of options ...

From the look of other implementations (angular, react, vue, etc), I think we better keep select as select and use menu for inline list of options. I can go ahead with `ux-menu` if needed to make it easier for us to justify

EisenbergEffect:
Demo is great! I played with a number of things, including custom elements for the option, keyboard, etc. Very nice work.

bigopon:
Cool, I didn't try with custom element, but nice to have you tested. The focus/blur part of select element always scares me so I tried to do it right first. Still need some more tweaks but it's not a disaster now, at least. What do you think about

> multiple should be a simiple list of options that you can select from, similar to the multiple select html control

bigopon:
From the look of [menu spec](https://material.io/guidelines/components/menus.html#menus-specs), I think `ux-select` is for our creativity that leverages aurelia templating power, while menu will be use as common case similar to native `<select>`/ `<option>` combo (The spec doesn't even have select !). The menu combo will be something like this

```html
  <ux-menu>
    <ux-menu-item icon="redo">Redo</ux-menu-item>
    <ux-menu-item icon="undo">Undo</ux-menu-item>
    <ux-menu-item icon="copy">Copy</ux-menu-item>
    <ux-menu-item icon="paste">Paste</ux-menu-item>
    <ux-menu-item icon="insert">Insert</ux-menu-item>
  </ux-menu>
```

@EisenbergEffect @ZHollingshead What do you think

EisenbergEffect:
Sounds good!

ZHollingshead:
What is left to do on this?

bigopon:
@ZHollingshead When looking at the spec of `menu`, I think we should delegate `select/option` to `ux-menu`. So I can't decide what to do next for this. All basic I could think of is done. Original intention is to have `ux-select` as both native `select/option` and advanced `dropdown`, supporting custom element, option template.

Other implementers did this quite differently, so I'm not sure.

ZHollingshead:
I think we should still have a simple ux-select component. But are you saying that it should use ux-menu for its dropdown?

ZHollingshead:
From a UX standpoint, as far as using the component itself, I think this is probably my favorite select input I have seen to date. http://www.material-ui.com/#/components/select-field

bigopon:
The reason I'm saying that is `material-ui` select implementation doesn't strictly follow spec, and I didn't know what our intention of doing this part is. example: select list of spec is designed to start where the current option is, while react version is top down. ([angular version for comparison](https://material.angular.io/components/select/overview))

ZHollingshead:
A select input component does not actually exist in the Material Spec. The closest we have is the Menu component as you mentioned, of which there are examples of it being used to make a pseudo select component, but not one that is explicitly outlined in the specification. And not every component we make will exist in the Material Specs, we also have IOS specs as well. Looking at your examples, you are spot on with this component in my opinion.

bigopon:
Disabled, enable scenario needs to be revised as it's not easy to use atm, also options grouping. I'm thinking of using 2 - 3 `replaceable/part` where user can have total control over the option template. Currently only content of a `ux-option` is override-able

ZHollingshead:
That sounds good to me.

bigopon:
close in favor of #145

PR #141, opened 2018-01-06
feat(AureliaUx): Use event for two ways binding

### LINKED ISSUE #140: Ux components & user input subscription
At the moment, we declared a `value` field with `@bindable({ defaultBindingMode: bindingMode.twoWay })` for user input (`ux-input`, `ux-textarea`, `ux-select` [i'm doing this]). I think we should follow the way we subscribe to built-in inputs: dispatch `change` event to notify subscribers. To do this, `Ux-core` needs to provide a method to intercept `determineDefaultBindingMode`  at [here](https://github.com/aurelia/templating-binding/blob/master/src/syntax-interpreter.js#L40) for subscribing to change event instead of binding directly to view model field.

This enable more fine-grained control on how to notify subscribers, benefits: `ux-select`, at least. I feel like declaring `@bindable({ defaultBindingMode: bindingMode.twoWay })` should be declared on application level components, not component level lib, plus the async nature of `bindable` will be problematic in the future I expect.

What do you think @EisenbergEffect @jdanyow @ZHollingshead

bigopon:
Changes:

### core

- requires https://github.com/aurelia/binding/pull/656 for observer typings
- adds `booleanAttr` binding behavior to get / set attribute & aria for boolean props
- takes dependency on `aurelia-templating-binding` (now intercepts `SyntaxInterpreter` class)
- adds new configuration API:

  ```ts
    registerUxElementConfig(config: {
      tagName: string
      properties: {
        [propertyName: string]: {
          defaultBindingMode: bindingMode,
          getObserver(
            obj: Element,
            propertyName: string,
            observerLocator: ObserverLocator
          ): InternalPropertyObserver
        }
      }
    });
  ```

    Example:

  ```js
    ux.registerUxElementConfig({
      tagName: 'ux-checkbox',
      properties: {
        checked: {
          defaultBindingMode: bindingMode.twoWay,
          getObserver(element, propertyName, observerLocator) {
            return new CheckedObserver(element, uxCheckboxChangeHandler, observerLocator);
          }
        }
      }
    });

    const uxCheckboxChangeHandler = {
      subscribe(target, callbackOrListener) {
        target.addEventListener('change', callbackOrListener, false);

        return function() {
          target.removeEventListener('change', callbackOrListener, false);
        };
      }
    };
  ```

----

#### overall goal for `<ux-input/>`, `<ux-textarea/>` and probably `<ux-select/>`, `<ux-checkbox/>`, `<ux-radio/>`:

-  will all notify their subscribers via `change` event.
- will not support `containerless`, as it relies on element dispatching event to notify subscribers. Can support aliasing via `as-element`.
- will mimic built-in controls by defining getter/setter on their elements, pointing to underlying view model values

#### `<ux-input/>`

- `value` is now real number when type is `number`. Thus empty string, input will give value `null`. `NaN` is also coerced into `null`.
- `null` / `undefined` are better handled, as they are kept as is instead of coerced into empty string.
- generally values as `null`/ `undefined` are kept as is, converted to number if type `number`, `string` if other types.

#### `<ux-checkbox/>`

- behaves in the same way with native `<input type="checkbox"/>`, add `indeterminate`

#### `<ux-textarea/>`

- generally values as `null`/ `undefined` are kept as is, converted `string` otherwise

---------

## TODO

- [x] Cleanup `<ux-checkbox/>`
- [x] Refactor `<ux-radio />`
  - [x] Add indeterminate
- [x] Refactor `<ux-textarea />`

----

* resolves #140


bigopon:
This is for the reason I mentioned in #140

bigopon:
@ZHollingshead I refactor `<ux-input/>` with the changes, please pull and give it a shot. 

`<ux-input/>` now has `rawValue: string` to handle internal value and `processRawValue` to handle `type = number`

bigopon:
@ZHollingshead @EisenbergEffect I updated the PR with some high level design decision. Would be nice if you could give some feedback based on new input

packages/checkbox/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
 export { UxCheckboxTheme } from './ux-checkbox-theme';
 
 export function configure(config: FrameworkConfiguration) {
+  config.container.get(AureliaUX).registerUxElementConfig(uxCheckBoxConfig);
   config.globalResources([
     PLATFORM.moduleName('@aurelia-ux/checkbox/ux-checkbox')
   ]);
 }
+
+const uxCheckBoxConfig = {
```
ZHollingshead: We need to be sure to extend this configuration to the [`@aurelia-ux/components`](https://github.com/aurelia/ux/tree/master/packages/components) package.

packages/input/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
 export { UxInputTheme } from './ux-input-theme';
 
 export function configure(config: FrameworkConfiguration) {
+  config.container.get(AureliaUX).registerUxElementConfig(uxInputConfig);
   config.globalResources([
     PLATFORM.moduleName('@aurelia-ux/input/ux-input')
   ]);
 }
+
+const uxInputConfig = {
```
ZHollingshead: We need to be sure to extend this configuration to the [`@aurelia-ux/components`](https://github.com/aurelia/ux/tree/master/packages/components) package.

packages/radio/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
 export { UxRadioTheme } from './ux-radio-theme';
 
 export function configure(config: FrameworkConfiguration) {
+  config.container.get(AureliaUX).registerUxElementConfig(uxCheckBoxConfig);
   config.globalResources([
     PLATFORM.moduleName('@aurelia-ux/radio/ux-radio')
   ]);
 }
+
+const uxCheckBoxConfig = {
```
ZHollingshead: We need to be sure to extend this configuration to the [`@aurelia-ux/components`](https://github.com/aurelia/ux/tree/master/packages/components) package.

ZHollingshead: Shouldn't this be `uxRadioConfig` rather than `uxCheckBoxConfig`?

packages/switch/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
 export { UxSwitchTheme } from './ux-switch-theme';
 
 export function configure(config: FrameworkConfiguration) {
+  config.container.get(AureliaUX).registerUxElementConfig(uxCheckBoxConfig);
   config.globalResources([
     PLATFORM.moduleName('@aurelia-ux/switch/ux-switch')
   ]);
 }
+
+const uxCheckBoxConfig = {
```
ZHollingshead: We need to be sure to extend this configuration to the [`@aurelia-ux/components`](https://github.com/aurelia/ux/tree/master/packages/components) package.

ZHollingshead: Shouldn't this be `uxSwitchConfig` instead?

packages/textarea/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
 export { UxTextareaTheme } from './ux-textarea-theme';
 
 export function configure(config: FrameworkConfiguration) {
+  config.container.get(AureliaUX).registerUxElementConfig(uxTextAreaConfig);
   config.globalResources([
     PLATFORM.moduleName('@aurelia-ux/textarea/ux-textarea')
   ]);
 }
+
+const uxTextAreaConfig = {
```
ZHollingshead: We need to be sure to extend this configuration to the [`@aurelia-ux/components`](https://github.com/aurelia/ux/tree/master/packages/components) package.

packages/radio/src/index.ts
```diff
@@ -1,9 +1,33 @@
...
+    checked: {
+      defaultBindingMode: bindingMode.twoWay,
+      getObserver(element: Element, _: string, observerLocator: ObserverLocator) {
+        return new CheckedObserver(element, uxCheckboxChangeHandler, observerLocator);
+      }
+    }
+  }
+};
+
+const uxCheckboxChangeHandler = {
```
ZHollingshead: Shoudn't this be `uxRadioChangeHandler`?

packages/radio/src/ux-radio.ts
```diff
@@ -1,144 +1,132 @@
...
+  PaperRipple,
+  normalizeBooleanAttribute,
+  linkProperty
+} from '@aurelia-ux/core';
 import { UxRadioTheme } from './ux-radio-theme';
+import { ElementEvents, DOM } from 'aurelia-framework';
 
 const theme = new UxRadioTheme();
 
+export interface UxRadioElement extends HTMLElement {
```
ZHollingshead: Should we export this in the index file?

packages/switch/src/ux-switch.ts
```diff
@@ -1,19 +1,31 @@
...
+import { computedFrom, observable } from 'aurelia-binding';
 import { inject } from 'aurelia-dependency-injection';
-import { StyleEngine, UxComponent, PaperRipple, normalizeBooleanAttribute } from '@aurelia-ux/core';
+import { StyleEngine, UxComponent, PaperRipple, normalizeBooleanAttribute, linkProperty } from '@aurelia-ux/core';
 import { UxSwitchTheme } from './ux-switch-theme';
+import { DOM, ElementEvents } from 'aurelia-framework';
 
 const theme = new UxSwitchTheme();
 
+export interface UxSwitchElement extends HTMLElement {
```
ZHollingshead: Should we export this in the index file?

bigopon:
@ZHollingshead All above done. The last 2 commits:

- Exports more in each component `index.ts`
- Exports more in `@aurelia-ux/components/index.ts`
- Renames `Textarea` to `TextArea` (please revert / override if it was intentional)
- Add temporary exports to `aurelia-binding` while waiting for aurelia/binding#656

For `@aurelia-ux/components/index`, calling configuration should do the job, maybe we don't have to do anything in there.

ZHollingshead:
This looks good. Will wait to merge, there is a fix going out for the `aurelia-ux/core` package and then we can merge this in.

PR #148, opened 2018-02-11
WIP feat(Menu): ux-menu element

bigopon:
Initial work

ux-menu

Markup:

```html
<ux-menu open.bind='menuOpen' trigger.bind='menuButton'>
  <ux-menuitem click.delegate='zoomIn()' helper-text='ctrl + [+]'>Zoom In</ux-menuitem>
  <ux-menuitem click.delegate='zoomOut()' helper-text='ctrl + [-]'>Zoom Out</ux-menuitem>
  <ux-menuitem click.delegate='copy()' helper-text='ctrl + C'>Copy</ux-menuitem>
  <ux-menuitem click.delegate='paste()' helper-text='ctrl + V'>Paste</ux-menuitem>
  <ux-menuitem click.delegate='closeApp()' helper-text='alt + F4'>Close</ux-menuitem>
</ux-menu>
```

static img

image

bindings:

  - `open`  => default mode: two way. true = show menu ,false = hide
  - `trigger`: what element will be used to trigger the menu
  - ?? (what else here ?)

TODO:

  - [ ] disabled / enabled states
  - [ ] nested menu
  - [ ] drop up, drop right, drop left configurable
  - [ ] left icon , right icon
  - [ ] theming

### DIFF
--- a/packages/menu/def.d.ts
+++ b/packages/menu/def.d.ts
@@ -1,0 +1,4 @@
+declare module '*.html' {
+  const value: string;
+  export = value;
+}
--- a/packages/menu/index.ts
+++ b/packages/menu/index.ts
@@ -1,0 +1,11 @@
+import {
+  FrameworkConfiguration,
+  PLATFORM,
+} from 'aurelia-framework';
+
+export function configure(config: FrameworkConfiguration) {
+  config.globalResources([
+    PLATFORM.moduleName('./ux-menu'),
+    PLATFORM.moduleName('./ux-menuitem')
+  ]);
+}
--- a/packages/menu/ux-menu.css
+++ b/packages/menu/ux-menu.css
@@ -1,0 +1,71 @@
+.ux-menu {
+  display: block;
+}
+
+.ux-menu-anchor {
+  position: absolute;
+}
+
+.ux-menu-panel {
+  position: relative;
+  max-width: 300px;
+  max-height: 400px;
+  padding: 8px 0;
+  background-color: #fff;
+  box-shadow: 0 5px 5px -3px rgba(0, 0, 0, 0.2), 0 8px 10px 1px rgba(0, 0, 0, 0.14), 0 3px 14px 2px rgba(0, 0, 0, 0.12);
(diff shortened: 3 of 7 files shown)


EisenbergEffect:
@bigopon This is fantastic work! @ZHollingshead Can you provide some feedback, particularly on how we'd want to handle icons. Another things to consider, could we do something cool with keyboard shortcuts, particularly making them automatically adjust based on platform e.g. cmd vs. ctrl.

bigopon:
@EisenbergEffect how the API would be if we are to try recognise those help text ? I think they are a bit arbitrary to do

ZHollingshead:
So I am not sure how to handle Icons right now. Especially given that the `ux-icon` package is so heavy at the moment. The way I would suggest handling those is have a set container that can contain whatever icon is provided (enabling the user to use a different icon set if they wish)

Something like
```
<ux-menuitem>
  <span class="icon">
    <!-- Icon of choice here -->
  </span>
  text
</ux-menuitem>
```

ZHollingshead:
I pulled this down and took a look at it. I think it works pretty great as it is. I think it is too rigid as it is. I think the menu item itself should be able to be set using a class as well as an element. That way I could for example have an anchor element that is a link as a menu item.

```html
<ux-menu open.bind='menuOpen' trigger.bind='menuButton'>
  <ux-menuitem click.delegate='zoomIn()' helper-text='ctrl + [+]'>Zoom In</ux-menuitem>
  <ux-menuitem click.delegate='zoomOut()' helper-text='ctrl + [-]'>Zoom Out</ux-menuitem>
  <a class="ux-menuitem" href="/${someId}/users">View Users</a>
</ux-menu>
```

Another idea I think would be neat is being able to simply target a button that will toggle the menu rather than having to toggle through a variable.

bigopon:
Agree that menu should be able to be triggered from an anchor button, I left it in initial markup as a todo. I think this can be continued after we adjust the build setup and initialization of the whole lib. I kind of want to wait for main PRs in templating before progressing this.

Issue #62, opened 2017-04-17
add karma testing


ZHollingshead:
Test commands currently will not work since there is no test directory or entry point for Karma.

bigopon:
@ZHollingshead Resolved. Thanks to @fkleuver 👍

PR #151, opened 2018-03-03
chore(test): switch to karma-webpack, make karma.conf typescript

fkleuver:
In part to address #62, I've made these changes to the test configuration:

-  Switch from requirejs to webpack
I believe this offers a more streamlined approach and make it easier to add unit and integration tests for components in the future
- Convert karma.conf.js to typescript
Have more information when working with the configuration, removes the need for comments to describe what each option does
- Add test coverage reporting
Output goes to /coverage folder and gives an accurate view of the code coverage of each tested component, making it easier to "figure out what tests to write"

I think overall this simplifies the test setup and speeds up the test runner as well.


fkleuver:
It seems like components would still be developed whilst having the entire repository open in the editor. This means that they can reuse some of the testing infrastructure that sits in the root, and they don't need a full-blown copy of the configs.

I see some alternative ways to do this but ultimately it's a trade-off between component autonomy and code duplication. I simply picked an approach that feels like a fair balance between the two, but it could be changed towards either direction still.

Settings up the tests with the correct karma/webpack/typescript configuration and some reuse was fairly straightforward. However the type checks have been giving me problems from start to finish - `lerna run build` wouldn't even work *before* I made any changes. The tests themselves work, but the components won't quite load/build. Only `core` seems to "work"?

ZHollingshead:
I think some of the builds may be broken on the master branch. It might not be your code at all, I am not sure which commit caused the break so I need to look more into that.

fkleuver:
Yeah it's definitely the build and I know what's causing it. Some files were added in [this commit](https://github.com/aurelia/ux/commit/5bf16cec07474a4018063afa09b533e3d6303398) by @bigopon but the `dist` are not yet updated to include those additions. You can see for example `core/components/link-property.ts` but there's no corresponding `dist/[output]/components/link-property.js/.d.ts`. Other components have started using them, and lerna sets up references to the `dist` folders and as such they cannot be resolved.

I'm not entirely sure how lerna works, but I'd say it should either be running the builds in the correct order (and there must be no circular dependencies) so that the latest stuff is always referenced, or it needs to setup references to the sources rather than the dist folders. Either that, or there need to be intermediary `dist` commits when stuff is added so lerna can grab those.

PR #157, opened 2018-03-08
chore(test): switch to karma-webpack, make karma.conf typescript

fkleuver:
Applies the same updates to `ux-select` from PR #151


fkleuver:
Made a mistake there with a force-push but recovered it again. Also added the updated webpack v4 types btw

PR #173, opened 2018-05-31
feat(grid): add IE11 polyfill

ben-girardet:
Here is an attempt to make ux-grid compatible with IE11 browsers.

This polyfill can be included in any project with 

```
import '@aurelia-ux/grid/ux-grid-ie11-polyfill';
``` 

It polyfills the rows and columns concepts, using prefixed -ms- CSS and supports responsive behavior. 

It doesn't cover the ordering concept.

### DIFF
--- a/packages/grid/src/ux-grid-ie11-polyfill.ts
+++ b/packages/grid/src/ux-grid-ie11-polyfill.ts
@@ -1,0 +1,96 @@
+let w: any = (window as any);
+let d: any = (document as any);
+// IE detection from: https://stackoverflow.com/questions/21825157/internet-explorer-11-detection
+let isIE11 = !!w.MSInputMethodContext && !!d.documentMode;
+
+if (isIE11) {
+  window.addEventListener('resize', () => {
+    fixUxGrids(true);
+  });
+}
+
+// DOM observer from: https://stackoverflow.com/a/14570614/437725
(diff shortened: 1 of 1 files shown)


ZHollingshead:
Does this cover the case where a grid can have a different number of columns?

i.e.

```
<ux-grid columns="6">
</ux-grid>
```

ben-girardet:
No it doesn't (yet) but it's possible to make it work in this case as well. I was not aware that ux-grid supported a columns attribute.

ZHollingshead:
Indeed. That is one thing that sets it apart from many grids, you can change the number of columns.

ben-girardet:
I’ve updated the pull request to support the colums attribute. It takes into account the original value given to the ux-grid component. 

What I haven’t been able to do is make the polyfill able to adjust when bindable properties are changed during runtime (such as columns or xs, sm, md, lg, xl attributes). I don’t know how to make this code able to observe these bindables.

iocast:
@ben-girardet I just tested your IE11 polyfill in IE11 and Edge. In Edge it polyfills the grid with `-ms` prefix. When I change these lines
```javascript
observeDOM( document, function() {
  fixUxGrids();
});
```
to
```javascript
if (isIE11) {
  observeDOM( document, function() {
    fixUxGrids();
  });
}
```
it works.

PR #175, opened 2018-06-22
refactor(All): move build to rollup, update core deps

### LINKED ISSUE #162: Switch all component build to rollup
With #147 in place, and incoming new way to register global resources via https://github.com/aurelia/framework/pull/858 , it's now possible to distribute every ux package in a single file., and it's more desirable to to do, for build pref, runtime startup perf, size reasons.

Rollup is a good bundler for this, and it will also make the build setup very simple.

 Thoughts ? @ZHollingshead @EisenbergEffect

bigopon:
Follow up the release of `aurelia-templating@1.8.0-rc.1` & `aurelia-framework@1.3.0-rc.2`, we can register global resources with direct view model references. Now ux elements as a single bundle file, like following pic

image

The entry `configure` will look like this:

```ts
import {
  FrameworkConfiguration,
  bindingMode,
} from 'aurelia-framework';

import { ValueAttributeObserver, EventSubscriber } from 'aurelia-binding';

import { AureliaUX } from '@aurelia-ux/core';
import { UxSelect } from './ux-select';
import { UxOptGroup } from './ux-optgroup';
import { UxOption } from './ux-option';

export { UxOption, UxOptionElement } from './ux-option';
export { UxOptGroup, UxOptGroupElement } from './ux-optgroup';
export { UxSelect, UxSelectElement } from './ux-select';
export { UxSelectTheme } from './ux-select-theme';

  config.container.get(AureliaUX).registerUxElementConfig(uxSelectConfig);
  config.globalResources([
    UxSelect,
    UxOptGroup,
    UxOption,
  ]);
}

const uxSelectConfig = {
  tagName: 'ux-select',
  properties: {
    value: {
      defaultBindingMode: bindingMode.twoWay,
      getObserver(element: Element, _: string) {
        return new ValueAttributeObserver(element, 'value', new EventSubscriber(['change']));
      }
    }
  }
};
```

This can be applied for all elements, which should help reduce build time too.

@EisenbergEffect @ZHollingshead I not sure how to update jspm dependencies properly.

resolves #162 

TODO list:

- [x] Button https://github.com/aurelia/ux/pull/175/commits/c4a82970593bf55b2d1a8d65e9a7bd913fec6e37
- [x] Card https://github.com/aurelia/ux/pull/175/commits/9c20ebf3a0b9b4682e01b539898c087d91afdb5c
- [x] Checkbox https://github.com/aurelia/ux/pull/175/commits/3f3ae9a9ce2e980f6fe8dc5824636bea6e561f3f
- [x] ChipInput https://github.com/aurelia/ux/pull/175/commits/ff825f80222fa8d71c840a65fc878f286f6545b1
- [x] Components
- [x] Core https://github.com/aurelia/ux/pull/175/commits/c114d781d3f8306aae6c18cdd22e0343c58cabe8
- [x] Datepicker https://github.com/aurelia/ux/pull/175/commits/e2273a4453dbd1057d9caf2cbaa0e306c0a8676f
- [x] Form https://github.com/aurelia/ux/pull/175/commits/6adda495dae06f8feebaa24bf8268876bda46f79
- [x] Grid https://github.com/aurelia/ux/pull/175/commits/b4c8bb46ad2001b11b726e85068d9d524bdc0b75
- [x] Icons **Renamed to __Icon__** https://github.com/aurelia/ux/pull/175/commits/935254544c6eff034972fd3f9d9bc55da976903a
- [x] Input https://github.com/aurelia/ux/pull/175/commits/12b61a6a86e40e039ae3610d6b95306a2224d4b6
- [x] Input-info https://github.com/aurelia/ux/pull/175/commits/ce0d803866bcd5cf1da4ee658d1e80c50b2e503e
- [x] List https://github.com/aurelia/ux/pull/175/commits/31ebac018335ed164096fac6f0595a972145d966
- [x] Radio https://github.com/aurelia/ux/pull/175/commits/62f921540f7e0c27163d70d204d00305186f3557
- [x] Select https://github.com/aurelia/ux/pull/175/commits/b8ccff4f5e93b2537e85d93767e8c17d04b4a4e8
- [x] Switch https://github.com/aurelia/ux/pull/175/commits/6025ed0961aff95ecc9b58d12b9ffa82ce7c2df9
- [x] Textarea https://github.com/aurelia/ux/pull/175/commits/a41573f6fa10fb14c778c7ed0f9a7c166559007a


bigopon:
@ZHollingshead Sorry Didn't state clearly WIP, I'm not sure if I should move wholesale to rollup, or one or few first and the rest later.

bigopon:
@ZHollingshead @EisenbergEffect  It would be also nice if we don't have to duplicate `css` files for each dist format, instead copy all css into a folder name `css`

ZHollingshead:
@bigopon Oh, whoops. Yeah I will get in and start testing.

I would be okay with doing it all at once if it works.

On the folder structure, originally you proposed an idea where we had the dist folder look like the following:
```
css/
types/
amd-index.js
commonjs-index.js
es2015-index.js
native-modules.js
```
If we do that, then we can just put the folders in the root one time yeah? What is required for that to work though, is that a situation where we need to configure loaders specially?

To be honest I am not sure how to update JSPM dependencies either. I am most experienced with Webpack at this point.

bigopon:
Maybe @jods4 can comment on how to have it work with webpack, without having to nested inside a folder.

EisenbergEffect:
I'd love to see this converted all at once, if possible :)

bigopon:
@EisenbergEffect @ZHollingshead That's pretty much it for the base work. Needs to update Webpack for html import in test, beside css test for all loaders to see if any issues.

packages/button/src/ux-button.ts
```diff
@@ -1,10 +1,12 @@
-import { customElement, bindable } from 'aurelia-templating';
+import { customElement, bindable, inlineView } from 'aurelia-templating';
 import { inject } from 'aurelia-dependency-injection';
 import { StyleEngine, UxComponent, PaperRipple, normalizeBooleanAttribute } from '@aurelia-ux/core';
 import { UxButtonTheme } from './ux-button-theme';
+import UX_BUTTON_VIEW from './ux-button.html';
```
ZHollingshead: Wouldn't we want the casing to be `UxButtonView` instead of `UX_BUTTON_VIEW` to be consistent?

packages/button/src/index.ts
```diff
@@ -1,9 +1,10 @@
...
+import { UxButton } from './ux-button';
+import css from './ux-button.css';
 
 export { UxButtonTheme } from './ux-button-theme';
 
 export function configure(config: FrameworkConfiguration) {
-  config.globalResources([
-    PLATFORM.moduleName('@aurelia-ux/button/ux-button')
-  ]);
+  DOM.injectStyles(css, undefined, undefined, 'ux-button-css');
```
ZHollingshead: So I am not sure how I feel about the style injection happening in the index page. This feels like it would be a better fit in the components view-model, either in a custom decorator or just in the constructor.

The ideal scenario would have been figuring out why the `require` tag is not working in Webpack when using rollup to move everything into a single file.

bigopon: We can copy the `.css` files over to dist to have it dynamically bundled. It will work fine, we just need to adjust the test to have those `css` file properly mapped

packages/button/src/ux-button.ts
bigopon: I was not sure about this, so I named it like a constant. Wouldn't be an issue if you want to proceed.

bigopon:
close as no longer relevant, resolved by #229

PR #225, opened 2019-11-20
fix(input): set input autocomplete attribute

ben-girardet:
As a first contribution I wanted to submit this small fix regarding the autocomplete attribute of the ux-input. It's a very small fix when two lines must be inverted.

My concern is that when I make this small change, build and submit a PR, suddenly it affects so many files and remove lots of /dist files.

Am I doing something wrong? I want to learn to do this correctly so I can continue contributing more. Thanks for your support.


bigopon:
@ben-girardet please wait a bit for abother PR of mine to fix some build issues. After that, @EisenbergEffect can release a new version and your pr can be checked againstneasier. The new build output will only have 1 js file, so you will be fine. Though we normally dont accept dist file in PR, so please dont commit them. Im pushing the build fix in an hour or so.

Finally, thanks for this PR, looking towards more as you evolve your apps ❤

ben-girardet:
Here is my update for this PR, without any `dist` files.

However now my problem is that even if I change two lines of code, the whole file is considered as changed, probably due to Line Ending issue or something but I tried to save with LF and CRLF, both cases seem to have the same problem.

Pictures below show the changes in VS Code. Any tips to commit the files without them being considered as totally changed ?

Capture d’écran 2019-11-21 à 15 58 11

Capture d’écran 2019-11-21 à 15 58 20

ben-girardet:
Okey I've been able to fix this. Seems that you guys are working on Windows with CRLF line endings. I tried on my Mac and Ubuntu, and both use LF line endings.

I've configured my GIT client with

```
git config --global core.autocrlf input
```

to fix this and now the PR committed files only contains the right edited lines.

Question though: would it not be better to have the GIT object database with files committed in LF in it ? Seems a little bit more standard. 

https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace

Anyway, for now I'm happy that it works :-)

bigopon:
Yes thats probably it. Though im not aure why my git isnt configured that way. Btw, should we use API toggleAttribute instead of 2 branches of if else? It may require a small polyfill though, but will result nin much cleaner code based

ben-girardet:
Are you on Windows ?

We should not use toggleAttribute as it is for Boolean attributes (https://developer.mozilla.org/fr/docs/Web/API/Element/toggleAttribute). The autocomplete is an attribute that accepts many strings (https://developer.mozilla.org/fr/docs/Web/HTML/Attributs/autocomplete).

PR #227, opened 2019-11-22
feat(input,textarea,datepicker,chip-input): focus api

ben-girardet:
Add a `focus()` and `blur()` public API to

* ux-input
* ux-textarea
* ux-datepicker
* ux-chip-input

Compatible with the Aurelia built-in `focus` custom attribute.

### DIFF
--- a/app/src/focus.ts
+++ b/app/src/focus.ts
@@ -1,0 +1,27 @@
+export interface StringTMap<T> { [key: string]: T; };
+export interface StringAnyMap extends StringTMap<any> {};
+export interface StringBooleanMap extends StringTMap<boolean> {};
+
+export class Focus {
+  private focused: StringBooleanMap = {
+    input: false,
+    textarea: false,
+    select: false,
+    datepicker: false,
+    chipInput: false
+  };
+
+  private viewModels: StringAnyMap = {}
+
+  public focus(component: string, usingCustomAttribute: boolean = false) {
+    if (usingCustomAttribute) {
+      this.focused[component] = true;
+    } else {
+      this.viewModels[component].focus();
+    }
+  }
+
+  public focusDatepicker() {
+    this.viewModels['datepicker'].focus(true);
+  }
+}
--- a/app/src/focus.ts
+++ b/app/src/focus.ts
@@ -1,6 +1,6 @@
-export interface StringTMap<T> { [key: string]: T; };
-export interface StringAnyMap extends StringTMap<any> {};
-export interface StringBooleanMap extends StringTMap<boolean> {};
+interface StringTMap<T> { [key: string]: T; }
+interface StringAnyMap extends StringTMap<any> {}
+interface StringBooleanMap extends StringTMap<boolean> {}
 
 export class Focus {
   private focused: StringBooleanMap = {
--- a/design-decisions.md
+++ b/design-decisions.md
@@ -25,7 +25,6 @@ Each component that implements the `focus()` and `blur()` API must have a distin
 
 * `ux-input` (done)
 * `ux-textarea` (done)
-* `ux-select` (work in progress)
 * `ux-datepicker` (done)
 * `ux-chip-input` (done)
 
@@ -33,6 +32,7 @@ Each component that implements the `focus()` and `blur()` API must have a distin
 
 *I am not sure if these components need to fully implement the focus() and blur() api. They do need to respond to <tab> keys and show a focused state. It seems to me, however, that they don't need to implement the focus() and blur() API as it doesn't make sense that we programatically want to focus a button or a checkbox?*
 
+* `ux-select` (tab works correctly)
 * `ux-checkbox` (tab works correctly)
 * `ux-radio` (tab works correctly)
 * `ux-switch` (tab works correctly)
--- a/design-decisions.md
+++ b/design-decisions.md
@@ -25,11 +25,16 @@ Each component that implements the `focus()` and `blur()` API must have a distin
 
 * `ux-input` (done)
 * `ux-textarea` (done)
-* `ux-button`
-* `ux-checkbox`
-* `ux-radio`
-* `ux-switch`
-* `ux-chip-input`
-* `ux-select`
-* `ux-datepicker`
-* `ux-slider`
+* `ux-select` (work in progress)
+* `ux-datepicker` (done)
+* `ux-chip-input` (done)
+
+### Components with input or action on click
+
+*I am not sure if these components need to fully implement the focus() and blur() api. They do need to respond to <tab> keys and show a focused state. It seems to me, however, that they don't need to implement the focus() and blur() API as it doesn't make sense that we programatically want to focus a button or a checkbox?*
+
+* `ux-checkbox` (tab works correctly)
+* `ux-radio` (tab works correctly)
+* `ux-switch` (tab works correctly)
+* `ux-slider` (tab works correctly)
+* `ux-button` (tab works correctly)
--- a/design-decisions.md
+++ b/design-decisions.md
@@ -1,0 +1,35 @@
+# `focus` attribute
+
+*You can contribute to the dicussion on this thread: https://github.com/aurelia/ux/issues/170*
+
+### API
+
+Inputable components should implement a `focus()` and `blur()` public API. Calling these methods must
+
+* display the component in its *focused* state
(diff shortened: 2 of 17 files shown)


app/src/focus.ts
```diff
@@ -0,0 +1,27 @@
+export interface StringTMap<T> { [key: string]: T; };
+export interface StringAnyMap extends StringTMap<any> {};
+export interface StringBooleanMap extends StringTMap<boolean> {};
+
+export class Focus {
```
bigopon: Can you help add some comment describing the purpose of this class?

design-decisions.md
```diff
@@ -0,0 +1,40 @@
...
+* display the component in its *focused* state
+* if text can be entered, it must focus the input tag, so that soft keyboard would open in mobile devices
+
+Also, if the inner HTML component receives focus (for exemple by the user hitting <TAB> key), the component must become in focused state.
+
+Since Aurelia has a built-in `focus` custom attribute, we expect Aurelia UX components to work seemlessly with this attribute. This constraints prevents us to use a `@bindable` for the focus attribute because the way Aurelia handles custom attributes conflicts with the bindable. The chosen solution is to add the `focus` and `blur` methods to the component's HTML element (through the custom prototype).
+
+### CSS
+
+When a component is in `focused` state, it must receive the CSS class `<comp-name>--focused`.
```
bigopon: pinging @EisenbergEffect @serifine for this change. At the moment, we are using `:focus` selector

app/src/focus.ts
ben-girardet: Hum... 

`StringAnyMap` and `StringBooleanMap` are only interfaces that I often use when I need typings in objects. They should not be exported.

design-decisions.md
```diff
@@ -0,0 +1,40 @@
...
+* `ux-input` (done)
+* `ux-textarea` (done)
+* `ux-datepicker` (done)
+* `ux-chip-input` (done)
+
+### Components with input or action on click
+
+*I am not sure if these components need to fully implement the focus() and blur() api. They do need to respond to <tab> keys and show a focused state. It seems to me, however, that they don't need to implement the focus() and blur() API as it doesn't make sense that we programatically want to focus a button or a checkbox?*
+
+* `ux-select` (tab works correctly)
```
bigopon: `ux-select` is a beast, in term of focus implementation, I think we can delay it if not critical

packages/chip-input/src/ux-chip-input.ts
```diff
@@ -181,4 +186,76 @@ export class UxChipInput implements UxComponent {
 
     this.styleEngine.applyTheme(newValue, this.element);
   }
+
+  public focusedChanged(focused: boolean) {
+    if (focused === true) {
```
bigopon: this is a good example of `toggle` api:
```html
this.element.classList.toggle('ux-chip-input--focused', focused);
```

design-decisions.md
ben-girardet: Would also like to have their input on this one. It seems that at the moment there is a mix of both `:focus` and `<comp-name>--focused`. I was thinking that we are trying to use more BEM conventions, hence this suggestion.

packages/chip-input/src/ux-chip-input.ts
```diff
@@ -181,4 +186,76 @@ export class UxChipInput implements UxComponent {
...
+  }
+
+  public focus() {
+    this.textbox.focus();
+  }
+
+  public blur() {
+    if (document.activeElement === this.textbox) {
+      this.textbox.blur();
+      this.addChip();
```
bigopon: adding chip on blur is quite a surprise, do we want this? at least it needs to be explicitly configured

packages/input/src/ux-input.ts
```diff
@@ -202,23 +202,38 @@ export class UxInput implements UxComponent {
...
+  }
 }
 
 function stopEvent(e: Event) {
   e.stopPropagation();
 }
 
 const getVm = <T>(_: any) => _.au.controller.viewModel as T;
-const uxInputElementProto = Object.create(HTMLElement.prototype, {
+const uxInputElementProto = Object.assign(
```
bigopon: no need for `Object.assign`

packages/textarea/src/ux-textarea.css
```diff
@@ -34,7 +34,7 @@
 .ux-textarea__bottom-border {
   height: 2px;
   background-color: #4043ff;
-  background-color: var(--aurelia-ux--chip-input-border-bottom-focus, var(--aurelia-ux--design-primary, #4043ff));
+  background-color: var(--aurelia-ux--textarea-border-bottom-focus, var(--aurelia-ux--design-primary, #4043ff));
```
bigopon: 👍

packages/textarea/src/ux-textarea.ts
```diff
@@ -125,6 +126,11 @@ export class UxTextArea implements UxComponent {
   }
 
   public rawValueChanged(rawValue: string) {
+    if (rawValue.length > 0) {
+      this.element.classList.add('ux-textarea--has-value');
```
bigopon: class list toggle

packages/chip-input/src/ux-chip-input.ts
ben-girardet: took this behavior from the previous blur listener (see detached() method on previous commit). For having used this component already in several projects I must say that I find it very practical from an end-user point of view.

When you say that it must be explicitly configured, do you think about a bindable to configure this as a component instance scope or from the plugin configure api to set it for all ux-chip-input in a project ?

bigopon: I see. yes it should be a bindable where we can toggle this behavior. Thanks for pointing out the existing behavior

ben-girardet: Will do.

As I like the `addChipOnBlur` behavior I would go for a default to `true` and allow user to opt-out. However, from your original remark you might suggest to have it `false` as default and require user to explicitly opt-in for it. Your thoughts ?

design-decisions.md
ben-girardet: This PR leaves the `ux-select` as is. I agree with you

packages/input/src/ux-input.ts
ben-girardet: Are you sure ? the focus() and blur() are added outside the Object.create() call. How would it look like without the `Object.assign` ?

packages/chip-input/src/ux-chip-input.ts
bigopon: Well, ux is used in prod, so we cannot flip it around. maybe default to `true` and have option to `false` it

packages/input/src/ux-input.ts
bigopon: Oh I see, is there any reason you don't put it inside the `Object.create` call?
```ts
Object.create(HTMLElement.prototype, {
  value: {
    get() {
      return getVm<UxInput>(this).getValue();
    },
    set(value: any) {
      getVm<UxInput>(this).setValue(value);
    }
  },
  focus: {
    value() {
      getVm<UxInput>(this).focused = true;
    }
  }
}),
```

ben-girardet: No reason, in fact I copied the code I saw in the related issue. Thanks for your exemple, I now removed `Object.assign()` in input, textarea, datepicker and chip-input.

app/src/focus.ts
ben-girardet: @bigopon maybe you were talking about the `Focus` class ? It's only the view-model for the `focus` custom element in the test app. I only wanted to have a dedicated code source to test the focus attribute and create a simple custom element for this.

ben-girardet:
@bigopon what's the status on this PR from your point of view ?

bigopon:
@ben-girardet as for the direction, I think it's good, as it enhances the custom element hosts to have similar API to native inputs. For the details of the focus behavior, I dont think we will have clear answer to all edge and corner cases, better let them with some consistency and then later draft out a better specs if we see any issues. For now, I think focusing on the underlying textbox is good enough. Let me know when you need a review for the code, as we don't have tests or specs, it will feel uneasy, but it should be ok if we continue the course of this until that point.

ben-girardet:
@bigopon You can go ahead with a code review. As you mention, it probably doesn't cover all cases, but from my experience (already using this code) it already greatly improve the focus api.

PR #229, opened 2019-11-23
Fix build ux

### LINKED ISSUE #228: Cannot convert object to primitive value with 0.17.0

* **Library Version:**
0.17.0

* **Operating System:**
OSX 10.x

* **Node Version:**
10.14.2

* **NPM Version:**
6.6.0

Webpack (from aurelia-cli)

* **Browser:**
Chrome XX

* **Language:**
TypeScript X.X

**Current behavior:**

When including the 0.17.0 version of Aurelia UX, the app builds normally but then the browser display the following error:

```
Possible Unhandled Promise Rejection: TypeError: Cannot convert object to primitive value
    at Object.createTemplateFromMarkup (aurelia-pal-browser.js?0627:476)
    at InlineViewStrategy.loadViewFactory (aurelia-templating.js?8628:642)
    at HtmlBehaviorResource.load (aurelia-templating.js?8628:4235)
    at eval (aurelia-framework.js?0682:276)
    at Array.map (<anonymous>)
    at loadBehaviors (aurelia-framework.js?0682:275)
    at eval (aurelia-framework.js?0682:395)
    at next (aurelia-framework.js?0682:162)
    at eval (index.js?07f7:57)
    at run (setImmediate.js?6017:40)
```

**How to reproduce**

```
au new <project-name> -u -s typescript,jest,vscode --install-deps npm
cd <project-name>
npm install @aurelia-ux/core @aurelia-ux/input
```

Adding to `main.js`:
```
.plugin(PLATFORM.moduleName('@aurelia-ux/core'))
.plugin(PLATFORM.moduleName('@aurelia-ux/input'))
```

```
au run
```

Error in browser: 
```
Possible Unhandled Promise Rejection: TypeError: Cannot convert object to primitive value
    at Object.createTemplateFromMarkup (aurelia-pal-browser.js?0627:476)
    at InlineViewStrategy.loadViewFactory (aurelia-templating.js?8628:642)
    at HtmlBehaviorResource.load (aurelia-templating.js?8628:4235)
    at eval (aurelia-framework.js?0682:276)
    at Array.map (<anonymous>)
    at loadBehaviors (aurelia-framework.js?0682:275)
    at eval (aurelia-framework.js?0682:395)
    at next (aurelia-framework.js?0682:162)
    at eval (index.js?07f7:57)
    at run (setImmediate.js?6017:40)
```

Could it be just a build issue from the 0.17.0 version ?

bigopon:
The type declaration in `html.d.ts` is different with what emitted by rollup html plugin (default es export). Tweak the types / import to align to the rollup plugin. Also fix the example app to adapt this change

closes #228


ben-girardet:
Will this close also #162  ?

bigopon:
Yes, it will. Im working some fixes locally, will push soon

bigopon:
@EisenbergEffect @ben-girardet I think i've finished it, this PR fixes html build issues, and will also fixes CSS bundling issue (atm, it's not giving any hints to webpack to bundle css, oops)

EisenbergEffect:
@bigopon Will this all continue to work correctly is someone is using the CLI bundler instead of Webpack?

bigopon:
I think so, because cli bundler does not have issue with module reference like webpack. Im using 2nd parameter of `@inlineView` decorator to add resources to a view. I think this is understood by the cli bundler,  cc @3cp

EisenbergEffect:
If we can get @3cp to do a quick review as well just in case, that would make me feel more comfortable. Thanks for putting the work in @bigopon

app/tsconfig.json
```diff
@@ -1,12 +1,17 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "target": "esnext",
+    "esModuleInterop": true,
```
3cp: Correct choice :-)

app/tsconfig.json
```diff
@@ -1,12 +1,17 @@
...
+    "allowSyntheticDefaultImports": true,
```
3cp: This is by default turned on when esModuleInterop is true.

packages/textarea/src/ux-textarea.ts
```diff
@@ -1,18 +1,21 @@
...
 export interface UxTextAreaElement extends HTMLElement {
   value: string;
 }
 
 @inject(Element, StyleEngine)
 @customElement('ux-textarea')
-@inlineView(VIEW)
+@inlineView(
+  VIEW,
+  [PLATFORM.moduleName('@aurelia-ux/textarea/ux-textarea.css')]
```
3cp: Using inlineView with css dep is fine for cli bundler, even when without `PLATFORM.moduleName` wrapper.

However, **this does not work for webpack**. This is a known issue, because this css dep is introduced from a js file, our default webpack config will use `style-loader` to inject it to HTML head (instead of leaving the responsibility to aurelia).

There are two issues in webpack.
1. doesn't support 2nd param in inlineView https://github.com/aurelia/webpack-plugin/issues/140
2. inline css dep through decorator doesn't work https://github.com/aurelia/cli/issues/911

The only solution is to avoid using inlineView, always use conventional html pair. You can only use inlineView if it doesn't introduce any css dep.

3cp: It doesn't matter the css dep is inside the template string of inlineView, or the 2nd param of inlineView. The webpack config see them same as the css dep is introduced from a js file.

app/tsconfig.json
bigopon: thanks, will clean this up

packages/textarea/src/ux-textarea.ts
bigopon: Thanks for the input here, I forgot about the behavior of style loader and the like

bigopon:
@EisenbergEffect it's ready to go. All builds have been reverted to multi-modules with modules referenced by strings. Also adjusted example app to ensure we can still develop with source code aliasing instead of having to build. @ben-girardet 

and huge thanks to @3cp for the links and reminder

ben-girardet:
thanks @bigopon looking forward to continue working on this lib.

PR #230, opened 2019-11-24
fix(switch): wrong aria attribute

### LINKED ISSUE #215: wrong attribute binding
https://github.com/aurelia/ux/blob/6b952e18c2d0268bdc6ea2d5cd7c3ad6b9316edf/packages/switch/src/ux-switch.html#L14

Shouldn't it be aria-disabled  instead aria-disabledB?

ben-girardet:
Fix #215


EisenbergEffect:
Nice catch!

PR #232, opened 2019-11-26
refactor(all): improve theming and CSS consistency

ben-girardet:
Please find here the WIP, targeting the `ux-button` component at the moment to improve theming and CSS consistency.

If you want to see it in practice, you can run the local app (detailed instruction in the main README)

```
cd app
npm ci
npm run dev
```

If you want to have quick look, here is a published dev version: https://ux.mintello.com
It should look like the GIF below and allow you to play with Aurelia UX core variables and `ux-button` theme properties and see the result live.

ezgif-1-88e670a2f35b

PR #235, opened 2020-02-02
feat(icon): icon map registration

ben-girardet:
@EisenbergEffect @bigopon 

As stated in https://github.com/aurelia/ux/pull/232#issuecomment-581082139 here is a proposal for a registration mechanism for `ux-icon`

Now the `ux-icon` contains no icons to start with but a `UxIconMap` static class that can handle icons registration.

Icons can be registered in the plugin configure call:

```ts
// main.ts
const icons = [
  {
    name: 'add',
    material: '...'
  },
  {
    name: 'remove',
    material: '...'
  },
];
.plugin(PLATFORM.moduleName('@aurelia-ux/icons'), icons)
```

or it can be registered later on: 

```ts
// app.ts
import { UxIconMap } from '@aurelia-ux/icons';

UxIconMap.registerIcons(icons);
```

I think it could be helpful to provide a few icon sets based on the current list. I've created a `sets` folder with two files: `minimal.ts` and `full.ts`.

Currently I've tested to import them in an application but I need to call them like: 

```ts
import { set as icons } from '@aurelia-ux/icons/dist/native-modules/sets/minimal';
```

I don't know how to build the icon sets so that we could import them like: 

```ts
import { set as icons } from '@aurelia-ux/icons/sets/minimal';
```

* Any feedbacks on this proposal ?
* Can someone help with the icons sets so that we could import them easily, and only those that we need in a project to prevent useless load ?


EisenbergEffect:
@bigopon may have some thoughts on improving the import situation.
In terms of the design, if possible, I'd prefer the icon map to be non-static. We can have a singleton instance injected into the icon element and we can create a configure method for the plugin that recieves the icons to register with the map at startup. Make sense?

ben-girardet:
Sure we can make it singleton. Happy to go this way. I am curious to understand the benefit of avoiding static classes. I guess that the purpose is to take advantage of the IoC?

If we make it singleton, what would be the way to register new icon sets outside of the plugin configuration ? Something like:

```ts
import { Container } from ‘aurelia-framework’;
import { UxIconMap } from ‘@aurelia-ux/icons’;

Container.instance.get(UxIconMap).registerIcons(...);
```

EisenbergEffect:
In general, we prefer to avoid static classes due to how they affect testing or potentially multiple aurelia app instances on the same page. So, we like everything to go through the IoC container so that it can be scoped to the Aurelia instance (if possible). Outside of plugin configuration, any class could inject the registry and register icons. Any any other plugin could do that as well.

```TypeScript
@inject(UxIconMap)
export class MyClass {
  constructor(iconMap) { ... }
} 
```

ben-girardet:
Makes total sense, thanks for clarifying the purpose.

ben-girardet:
@EisenbergEffect : https://github.com/aurelia/ux/pull/235/commits/9a6dedd40227f3ebeb3644fbeac5e9e3765a1606

packages/icons/src/index.ts
```diff
@@ -9,7 +9,7 @@ export { UxIconMap };
 
 export function configure(config: FrameworkConfiguration, icons?: Array<{name: string, material: string}>) {
   config.globalResources(PLATFORM.moduleName('./ux-icon'));
-  if (Array.isArray(icons) && icons.length > 0 && typeof icons[0].name === 'string' && typeof icons[0].material === 'string') {
-    UxIconMap.registerIcons(icons);
+  if (Array.isArray(icons) && icons.length > 0) {
+    Container.instance.get(UxIconMap).registerIcons(icons);
```
EisenbergEffect: Instead of using `Container.instance` we should use `config.container`. The `FrameworkConfig` instance will return the container instance for the current app, which is better to use in this case.

ben-girardet: Of course, using `Container` report the static issue to the Container class. Thanks for the feedback.

ben-girardet:
@EisenbergEffect fixed the Container issue.

@bigopon regarding the _import_ question above, I'm trying another solution: using `.json` files for the icons maps and placing them in the main package directory. This way they will end up at the root of the package and can be imported as

```
import fullSet from `@aurelia-icons/sets/full.json`
```

What do you think ?

ben-girardet:
Wanted to try to use the registration mechanism to bridge with another icon library. With FontAwesome it could look like this:

1. Register the component without icons

```ts
// main.ts
// ...
  .plugin(PLATFORM.moduleName('@aurelia-ux/icons'))
// ...
```

2. Import the FA JS icons and parse them
  
```ts
// app.ts

import { UxIconMap } from '@aurelia-ux/icons';
import { inject } from 'aurelia-framework';

import '@fortawesome/fontawesome-free/js/brands';
import '@fortawesome/fontawesome-free/js/regular';

const w = (window as any);

@inject(UxIconMap)
export class App {

  constructor(private uxIconMap: UxIconMap) {
    if (w.___FONT_AWESOME___ && w.___FONT_AWESOME___.styles) {
      type FabIcon = [Number, Number, Array<any>, string, string];
      for (let fKey in (window as any).___FONT_AWESOME___.styles) {
        let fIcons: {[key: string]: FabIcon} = (window as any).___FONT_AWESOME___.styles[fKey];
        let fSet: Array<{name: string, material: string}> = [];
        for (let iconKey in fIcons) {
          const icon = fIcons[iconKey];
          fSet.push({name: `fab_${iconKey}`, material: `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 ${icon[0]} ${icon[1]}"><path d="${icon[4]}"/></svg>`});
        }
        this.uxIconMap.registerIcons(fSet);
      }
    }
  }

}
```

End result:

icons


I find it cool :-)

ben-girardet:
@bigopon do you have any feedback/input regarding this PR and more specifically the `import` question for Icon Sets. Is it a good idea to have the icon sets as `JSON` files that we can import from the root directory ?

ben-girardet:
I've quickly put together a little tool to generate Icon Sets (JSON, but could be anything once we agree on the format). 

It currently generate sets from https://materialdesignicons.com/ (more sources can be added later).

See here: https://github.com/platform5/ux-icon-builder

EisenbergEffect:
This makes sense and it seems there's some precedent from other projects as well.

ben-girardet:
@EisenbergEffect @serifine @bigopon 

This PR is ready for final review.

packages/icons/src/ux-icon-map.ts
```diff
@@ -1,3848 +1,38 @@
...
-        name: 'toll',
-        material: '<svg viewBox="0 0 24 24"><path d="M15 4c-4.42 0-8 3.58-8 8s3.58 8 8 8 8-3.58 8-8-3.58-8-8-8zm0 14c-3.31 0-6-2.69-6-6s2.69-6 6-6 6 2.69 6 6-2.69 6-6 6zM3 12a5.99 5.99 0 0 1 4-5.65V4.26C3.55 5.15 1 8.27 1 12s2.55 6.85 6 7.74v-2.09A5.99 5.99 0 0 1 3 12z"></path></svg>'
-      },
-      {
-        name: 'touch_app',
-        material: '<svg viewBox="0 0 24 24"><path d="M9 11.24V7.5a2.5 2.5 0 0 1 5 0v3.74c1.21-.81 2-2.18 2-3.74C16 5.01 13.99 3 11.5 3S7 5.01 7 7.5c0 1.56.79 2.93 2 3.74zm9.84 4.63l-4.54-2.26c-.17-.07-.35-.11-.54-.11H13v-6c0-.83-.67-1.5-1.5-1.5S10 6.67 10 7.5v10.74l-3.43-.72c-.08-.01-.15-.03-.24-.03-.31 0-.59.13-.79.33l-.79.8 4.94 4.94c.27.27.65.44 1.06.44h6.79c.75 0 1.33-.55 1.44-1.28l.75-5.27c.01-.07.02-.14.02-.2 0-.62-.38-1.16-.91-1.38z"></path></svg>'
-      },
-      {
-        name: 'track_changes',
-        material: '<svg viewBox="0 0 24 24"><path d="M19.
```
EisenbergEffect: Could we name this `icon` instead so that the material design naming doesn't become part of the api itself? We'd like to do other design systems one day :)

ben-girardet:
@EisenbergEffect in fact I'm suggesting to go one step further. Based on my observation of FontAwesome JS icons, I discovered that they save them in an JSON Array like:

```
[
  [
    "icon-name",
    24, // view box width
    24, // view box height
    "..." // charCode
    "M19" // svg path
  ]
]
```

Because they also provide their sets in font format they also include a `charCode` value. In our case, because we are SVG based, we could do :

```
[
  [
    "icon-name",
    "M19", // svg path
    24, // optional view box width (default 24)
    24 // optional view box height (default 24)
  ]
]
```

With this we can reduce the JSON size a little by removing the `SVG` markup and only providing width/height for non-default sizes (such as when we import FontAwesome icons).

Thoughts ?

EisenbergEffect:
@ben-girardet That sounds great. Let's do it.
BTW Would you mind emailing me at rob at bluespire dot com ?

packages/icons/src/ux-icon-map.ts
```diff
@@ -1,24 +1,47 @@
...
-    if (Array.isArray(icons) && icons.length > 0 && typeof icons[0].name === 'string' && typeof icons[0].material === 'string') {
-      icons.map(icon => this.registerIcon(icon.name, icon.material));
+  public registerIcons(icons: Array<UxIconRegArray | UxIconRegObject>) {
+    if (Array.isArray(icons)) {
+      icons.map(icon => this.registerIcon(icon));
     }
   }
 
+  private buildSvg(icon: UxIconRegArray): string {
+    return `<svg viewBox=\"0 0 ${icon[2] || 24} ${icon[3] || 24}\">${icon[1]}</svg>`;
```
EisenbergEffect: Could we extract `24` into a class-level property so that the default size can be set externally if needed?

packages/icons/src/ux-icon-map.ts
```diff
@@ -1,24 +1,47 @@
...
+      return;
+    }
+    this.map[name] = svg;
   }
 
-  public registerIcons(icons: Array<UxIconReg>) {
-    if (Array.isArray(icons) && icons.length > 0 && typeof icons[0].name === 'string' && typeof icons[0].material === 'string') {
-      icons.map(icon => this.registerIcon(icon.name, icon.material));
+  public registerIcons(icons: Array<UxIconRegArray | UxIconRegObject>) {
+    if (Array.isArray(icons)) {
```
EisenbergEffect: Is this check needed?

ben-girardet: I placed this check because this function is called from the plugin config. When a user configure the plugin and pass an IconMap to the plugin it doesn’t have type checking (as far as I know). This test is a small guard to avoid a blocking error if an incorrect set is passed. 

However I think that if I test I should console a warning in the ‘else’ block.

Otherwise I could also remove the test... other ideas?

EisenbergEffect: I think we can leave it as is. That's fine.
Ok, do you feel it's ready to merge @ben-girardet ?

ben-girardet: @EisenbergEffect  Just want to extract the icon default in the class level property and then we can merge. We should also prepare a doc or info for the breaking change as the icon set will not be imported by default. What would be the best way to do this?

EisenbergEffect: If you want to add a section to the PR, I can include that in release notes.

ben-girardet:
This version do not automatically import any icons by default. From now on, you must decide on a per-project basis what icon sets you want to use. As a quick start, you can import the same icon sets that was previously shipped with the package by configuring the plugin as such in your `main.ts`

```ts
/* Add this line to import the full set into an `icons` variable */
import icons from '@aurelia-ux/icons/sets/full-array.min.json';

/* When you initilize the plugin, you can pass on this `icons` variable */
aurelia.use.plugin(PLATFORM.moduleName('@aurelia-ux/icons'), {icons: icons})

/* That's it ! */
```

Note for typescript users: make sure that your `tsconfig.json` is set to allow importing JSON files. This can be achieved by settings `"resolveJsonModule": true` in the `compilerOptions` section.

ben-girardet:
@EisenbergEffect I've updated the PR with a class-level (configurable) property for `defaultIconWidth` and `defaultIconHeight`.

I didn't know what you meant by adding a _section_ to the PR so I wrote a short note in my previous comment that could be added to the release note.

I think this PR is ready for merge.

bigopon:
Nice! I think we may also want to add another small example of how to fetch it dynamically before initializing the icon package, or after initialization as well. Folks may have valid reasons to do it lazily.

ben-girardet:
Good idea @bigopon I've added more exemples by [editing my comment](https://github.com/aurelia/ux/pull/235#issuecomment-585434051) above.

bigopon:
@ben-girardet in the step 3, where does `___FONT_AWESOME___` come from? And thanks for the comment, I think this will be a pretty good baseline for later doc improvement related to icons

ben-girardet:
> @ben-girardet in the step 3, where does `___FONT_AWESOME___` come from? And thanks for the comment, I think this will be a pretty good baseline for later doc improvement related to icons

good catch, forgot the `import` part of the Font Awesome packages. Updated my comment, and added a section to explain how to *get* and *see* the registered icons.

ben-girardet:
@EisenbergEffect I've rebased my branch and merge with this PR. I've added some lint fixes.

@bigopon do you know if the `npm run test` or `npm run doc` are supposed to work? I've tried to run them and get quite a few errors that I struggle to fix. I'm trying to understand if I broke something or if these tests and docs were already not working before hand ?

bigopon:
@ben-girardet for now, both of them may not work. If it's not blocking you, give me a few days and i'll resolve it to prepare for a more solid infra with regards to those. If you have any ideas, can you put it here for now?

ben-girardet:
@bigopon thanks. It's not blocking for now and I believe we can merge this PR as this.

In regards to tests and docs I must say that I'm not (yet) familiar with setting this up. Very looking forward to see how you can set it up for this lib and learn from then.

EisenbergEffect:
@ben-girardet With the merge of this PR, do you think we're ready for a release?

ben-girardet:
@EisenbergEffect I've tried to clone, build and run the dev app and all works fine.

My last question regards `/dist` files. In the PR I've only pushed the `/src`. Are you building the dist before the release or not ? Just asking as I'm very new with the process.

ben-girardet:
Just found a small bug in the icon package. Will submit a new PR with the fix in a moment...

EisenbergEffect:
Got the bug fix PR in. Regarding the `dist` folder, I'll build that as part of the release process. With that in mind, anything else we need to do before I release or are we ready to go?

ben-girardet:
Ready to go ! Thanks @EisenbergEffect

PR #237, opened 2020-02-23
fix(icon): avoid unset theme property

ben-girardet:
For when a size is provided but no theme


ben-girardet:
@EisenbergEffect this fixes the bug I mentioned in #235

PR #238, opened 2020-02-25
feat(icon): simplify loading full icon set

ben-girardet:
@EisenbergEffect @bigopon 

I figured that using the icon set in JSON could be a little more complicated than expected (esp. making sure Typescript load the JSON correctly). Also, it's not recommended to imports files from sub directories.

Here is a suggestion to use the Loader and specifically request the bundler to create a separate file for the icon set. With this in place it's a whole lot easier to keep compatibility with previous version by simply opting in for the full icon set at config as such: 

```ts
// main.ts
.plugin(PLATFORM.moduleName('@aurelia-ux/icons'), {loadFullSet: true})
```

What are your thoughts ?


packages/icons/src/ux-icon-map.ts
```diff
@@ -9,11 +12,29 @@ export interface UxIconRegObject {
...
+@inject(Loader)
 export class UxIconMap {
 
   public defaultIconWidth: number = 24;
   public defaultIconHeight: number = 24;
 
+  constructor(private loader: Loader) {}
+
+  public loadFullSet() {
+    PLATFORM.moduleName('./full', {chunk: 'full-icon-set'});
```
bigopon: this can just be
```ts

    PLATFORM.moduleName('./full', 'full-icon-set');
```
unless there's some specific reason for longer form syntax?

bigopon:
I think this is fine, as long as we don't introduce any issues with bundling. And it seems like it shouldn't.

packages/icons/src/ux-icon-map.ts
ben-girardet: good catch. came from previous trials with JSON loading. The longer form syntax is not necessary here.

ben-girardet:
Yeah, I don't have massive experience with bundling, but from my tests it works as it should, creating a different chunk loaded only if necessary.

ben-girardet:
Closing in favor of #239

PR #239, opened 2020-02-26
feat(icons): full icon set available in .js

ben-girardet:
This simplify the import from web projects as it doesn’t require support for json import in tsconfig


ben-girardet:
@bigopon This replaces #238 in order to avoid issues with bundler but still simplifying the imports by providing a `.js` version of the icon set. 

Would you mind having a look for review ?

PR #242, opened 2020-03-19
fix: use esModuleInterop to fix moment rexport

3cp:
chore: enable esm dist file for various bundlers

More details in https://discourse.aurelia.io/t/aurelia-ux-datepicker-0-18-0-systemjs-cli/3307/2?u=huochunpeng

As you can see the esModuleInterop only affected commonjs and amd build for moment-rexports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ Check out the [showcase application](https://github.com/aurelia/app-ux-showcase)
 
 **Before The First Build**
 
-`npm ci`: installs dependencies for the base Aurelia UX project
-`npm run bootstrap`: sets up a symlink between all of the packages in the monorepo
-`npm run build`: builds all of the mono repo projects.
+* `npm ci` installs dependencies for the base Aurelia UX project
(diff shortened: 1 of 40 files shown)


3cp:
Although I could not run tests because karma-wepback setup does have proper css loader. I guess this is a known issue.

ben-girardet:
Thanks a lot for this PR @3cp 

I've reviewed and tested so from my opinion it's ready for merge. @EisenbergEffect 

My only question is about the /dist files. Should they be removed before merge ?

3cp:
As long as the dist files is generated when publish new version, it can be left out in this PR. Let me know if you want me to remove the last commit.

ben-girardet:
They will be generated for each new version so you can safely remove them. When I started working on the project we told me this is common practice, hence my comment. I would think that it doesn't cause any harm to leave them but it keeps the PR easier to review when they are left out.

3cp:
Done.

ben-girardet:
Fantastic! Ready to merge then

PR #246, opened 2020-03-23
WIP: Drawer and ModalService

ben-girardet:
@EisenbergEffect @bigopon @serifine 

Here is a draft for the `ux-drawer` component and a `ModalService`.

At the moment I have focused the work towards a _drawer_ component, but to be honest it works also as _dialog_. Basically the drawer can be set at `position`:

- `left`, `right`, `bottom` and `top` => looks like a drawer
- `center` => looks like a dialog

The `ModalService` is called as such because the intention is to use the same service across different components that would need to act as modal. This helps with keeping track of zIndex and properly layer components if several modals must be opened in a row. It is currently working only for drawers, but we can add support for other components as they come.

Drawers can be used either inline or with the `ModalService`. If they are used inline they don't need the service at all. I've set up a few examples in the dev app.

**Inline use**

```html
<!-- Check out examples and code source for more -->

<ux-drawer if.bind="anyVarFromTheHostingVM">
  <!-- any content here that will be included in the slot -->
  <!-- I recommend using a ux-card as they fit well the purpose of this kind of surface -->
  <!-- BTW: I've set up a few CSS for when cards are used as drawer/dialog markup -->
</ux-drawer>
```

**Service use**

```ts
import { ModalService } from '@aurelia-ux/drawer';
import { inject } from 'aurelia-framework';
import { Hello } from './hello';

@inject(ModalService)
export class App {
  constructor(private modalService: ModalService) {}

  public openDrawer() {
    const drawer = this.modalService.open({viewModel: Hello, position: 'bottom', theme: {/*... */}});
    drawer.whenClosed().then((result) => {
      if (!result.wasCancelled) {
        console.log('output', result.output);
      }
    })
  }
}


```

For the `ModalService` I've tried to stay as close as possible to the current `DialogService` and `DialogController` api from `ux-dialog` plugin (except that the service handles both creating the drawer with `open()` and the `cancel()` and `ok()` apis from the included VM. 

The `.open()` api accepts a `viewModel`, `view` or both (+ some more options). It returns a `UxDrawer` instance with an additional `whenClosed()` promise similar to current `DialogService`. 

**Known limitations**

When using the `ModalService`, there are two limitations compared to the `ux-dialog`

1. missing `canActivate()` lifecycle hook. Because I rely on `<compose>` to build the VM/V, then I can't can `canActivate` from the service and then compose do not do that either. If this would be an important feature then we should not rely on `<compose>` and create the composition from the service itself
2. again, due to the `<compose>` usage, the included view cannot have any classes, attributes or event listener on the `<template>` tag.

This is a first draft that I want to publish here for early feedbacks.

I've setup my current showcase link as a demo of this PR: https://ux.mintello.com

Here is a quick gif to see the result:

_Various Positions_
drawer1

___
___
___
___
___
___
___
___
___

_Mobile_
drawer2

___
___
___
___
___
___
___
___
___

_Go Crazy_
drawer3


EisenbergEffect:
Thanks @ben-girardet This is looking nice. I admit to having a lot of fun playing with the demo :)
Unless I missed it, I think we need to add some stuff around accessibility. That's the only thing that jumped out at me in the code. When playing with the demo, I did see some odd behavior when opening overlapping drawers. I didn't dig in to see what was going on though.

ben-girardet:
Thanks @EisenbergEffect  for your feedbacks.

_Accessibility_: did you mean keyboard support or did you mean something else ? I pushed a commit with keyboard support now.

_Odd behavior_: you're right, I've also discovered that if we open one or several "crazy" drawer and then close them... and after that we open an inline drawer => then there will be as many inline drawer elements attached to DOM than we previously opened "crazy" drawers.

I don't exactly know what is going on here. I understand that the issue comes because the crazy drawer attaches the same module again (containing the inline drawer) and it probably means that the inline drawer is somehow _not properly detached_ when the module go away.

Looking forward to get more feedbacks. Especially about these questions:

* The way I move the attached element to its host (default to body)
* The way the service creates the drawer, with the use of `<compose>`, a temporary `bindingContext` for passing the options as bindables and `enhance()`

EisenbergEffect:
@ben-girardet I actually like the way you used compose and enhance together. Very clever and the code seems much cleaner than I remember some of the old dialog code being (though I haven't looked at that in a while).

In terms of accessibility, yes keyboard support, but also there may be some aria attributes we need to add to the backdrop and modal containers.

packages/drawer/src/ux-drawer.html
```diff
@@ -1,9 +1,16 @@
...
+  class="ux-drawer ux-drawer--${computedType} ux-drawer--${viewportType} ux-drawer--${position} ${animate ? 'ux-drawer--animate':''} ${showed ? 'ux-drawer--opened':''}">
   <require from="./ux-drawer.css"></require>
   <div class="ux-drawer__overlay" ref="overlayElement">
   </div>
-  <div class="ux-drawer__content-wrapper" ref="contentWrapperElement" click.delegate="overlayClick($event)">
-    <div class="ux-drawer__content" ref="contentElement" click.delegate="stop($event)">
+  <div 
+    role="${role || 'dialog'}"
+    aria-labelledby="ariaLabelledBy"
+    aria-describedby="ariaDescribedBy"
```
ben-girardet: @EisenbergEffect

For the `role` I've added a bindable that defaults to `dialog` (user can set it to `alertdialog` which is also valid for some cases.

Regarding `arial-labelledby` and `aria-describedby` the are content specific. Therefore I leave them here as bindable values. Users can set them according to their usage.

Once the core features of this drawer/dialog component are set I would like to provide a few utility methods with a default template to quickly display dialogs such as alerts, prompt and confirm. I would also like to provide an empty template for dialogs and users who like to have the template ready. I'm planning on settings these templates with `ux-card` as I do now in the demo.

Why I say all that is that for these cases I can set the correct aria properties directly. But for all user custom implementation then I can't do it for them.

ben-girardet:
Updated with component to make it work even more like current `ux-dialog` plugin.

* Added a `restoreFocus` bindable method to bring back focus to the main view when modal closes. Automatically defaults to lastActiveElement before opening the modal. 
* Added `canDeactivate`, `detached` and `deactivate` lifecycle support for VM passed on to the drawer/dialog service. Identical API than current ux-dialog plugin
* Fixes call to `detached` on ux-drawer when closing service handled drawer
* Added `attach-focus-drawer` attribute

ben-girardet:
I've added this paragraph to the PR description:

> **Known limitations**
> 
> When using the `ModalService`, there are two limitations compared to the `ux-dialog`
> 
> 1. missing `canActivate()` lifecycle hook. Because I rely on `<compose>` to build the VM/V, then I can't can `canActivate` from the service and then compose do not do that either. If this would be an important feature then we should not rely on `<compose>` and create the composition from the service itself
> 2. again, due to the `<compose>` usage, the included view cannot have any classes, attributes or event listener on the `<template>` tag.

ben-girardet:
Quick note to say that I think it is valuable to explore how to call the `canActivate` hook. 

It is an important piece in the VM lifecycle. It is the only way to give users a chance to validate if the VM receives the right `model` to create and handle the modal.

packages/drawer/src/ux-drawer.html
EisenbergEffect: That makes sense and this all looks good. Thanks!

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  constructor(
+    private templatingEngine: TemplatingEngine, 
+    private eventAggregator: EventAggregator,
+    private defaultConfig: DefaultDrawerConfiguration) {
+
+  }
+
+  public addLayer(drawer: UxDrawer, bindingContext: ModalBindingContext) {
+    this.modalLayers.push({
```
bigopon: Passing the tip around:
```suggestion
    const layerCount = this.modalLayers.push({
      bindingContext: bindingContext,
      drawer: drawer
    });
    if (layerCount === 1) {
      this.setKeyListener();
    }
```

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+      drawer.detached();
+      const parent = drawer.element.parentNode;
+      if (!parent) { return; }
+      parent.removeChild(drawer.element);
+      this.eventAggregator.publish(`drawer-${drawerIndex}-resolve`, {
+        wasCancelled: false,
+        output: event.detail
+      });
+    }
+    return (drawer as UxDrawer & ModalServiceDrawer);
```
bigopon: nit: No need for parens here

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+      const parent = drawer.element.parentNode;
+      if (!parent) { return; }
+      parent.removeChild(drawer.element);
+      this.eventAggregator.publish(`drawer-${drawerIndex}-resolve`, {
+        wasCancelled: true,
+        output: null
+      });
+    }
+    bindingContext.ok = (event: CustomEvent) => {
+      drawer.detached();
```
bigopon: `detached` being called before the actual removal of the element is quite unexpected, compared with the rest. I think we can add another lifecycle here if needed, maybe `detaching`/`deactivating` @EisenbergEffect

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    (drawer as UxDrawer & ModalServiceDrawer).whenClosed = () => {
+      return whenClosed;
+    }
+    bindingContext.dismiss = () => {
+      drawer.detached();
+      const parent = drawer.element.parentNode;
+      if (!parent) { return; }
+      parent.removeChild(drawer.element);
+      this.eventAggregator.publish(`drawer-${drawerIndex}-resolve`, {
```
bigopon: this event aggregator will be potentially shared by many drawers, there need to be more information to identify which drawer published this event, or we just dispatch event on the host of this drawer

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (layer.bindingContext && layer.bindingContext.composeViewModel && layer.bindingContext.composeViewModel.currentViewModel) {
+      const vm = layer.bindingContext.composeViewModel.currentViewModel;
+      if (typeof vm.deactivate === 'function') {
+        await vm.deactivate.call(vm, result);
+      }
+    }
+    return;
+  }
+
+  public async cancel(drawer?: UxDrawer) {
```
bigopon: does this need to be `async`?

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+    
+    element.setAttribute('host.bind', 'false');
+    
+    if (options.theme) {
+      bindingContext.theme = options.theme;
+      element.setAttribute('theme.bind', `theme`);
+    }
+    const compose = document.createElement('compose');
+    compose.setAttribute('view-model.ref', 'composeViewModel');
+    element.innerHTML = compose.outerHTML;
```
bigopon: I think we can merge this line and the 2 above it into 1:
```ts
element.innerHTML = `<compose view-model.ref=composeViewModel>`;
```
it's shorter and easier to understand

packages/drawer/src/ux-modal-service.ts
```diff
@@ -0,0 +1,277 @@
...
+  public get zIndex() {
+    return this.startingZIndex + this.modalLayers.length;
+  }
+
+  public open(options: ModalServiceOptions): UxDrawer & ModalServiceDrawer {
+    // const defaultConfig = this.container.get(DefaultDrawerConfiguration);
+    options = Object.assign({}, this.defaultConfig, options);
+    this.drawerIndex++;
+    const bindingContext: ModalBindingContext = {};
+    if (!options.viewModel && !options.view) {
```
bigopon: maybe doing this check right after normalizing `options` in line 140, it's better and can help avoid 1 mutation of `drawerIndex`, less surprises during debugging

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  private setZindex() {
+    this.overlayElement.style.zIndex = `${this.modalService.zIndex}`;
+    this.contentWrapperElement.style.zIndex = `${this.modalService.zIndex}`;
+  }
+
+  private moveToHost() {
+    const host = this.getHost();
+    if (!host) { return };
```
bigopon: comment for the whole PR, a bit of a linting/nit picking: please dont do same line return, it's easy to miss & not consistent with the rest of our modules

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,302 @@
...
+    this.showed = false;
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        this.modalService.removeLayer(this);
+        this.hiding = false;
+        if (this.lastActiveElement && typeof this.restoreFocus === 'function') {
+          this.restoreFocus(this.lastActiveElement);
+        }
+        resolve();
+      }, duration + 10);
```
bigopon: can we have a small comment to explain this `+ 10`?

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  private show() {
+    if (this.showing && this.showed) {return;}
+    if(DOM.activeElement instanceof HTMLElement) {
+      this.lastActiveElement = DOM.activeElement;
+    }
+    this.showing = true;
+    this.modalService.addLayer(this, this.bindingContext);
+    this.setZindex();
+    this.taskQueue.queueTask(() => {
```
bigopon: why task instead of micro task? Please help add some comment for timing related code

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  public detached() {
+    if (this.host) {
+      this.removeFromHost();
+    }
+  }
+
+  private show() {
+    if (this.showing && this.showed) {return;}
+    if(DOM.activeElement instanceof HTMLElement) {
```
bigopon: if we are to use `DOM`, better use it everywhere, else we can use `document.activeElement` instead

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (this.host === '') {
+      this.host = 'body';
+    }
+  }
+
+  public overlayDismissChanged() {
+    if (!this.overlayDismiss && this.defaultConfig.overlayDismiss) {
+      this.overlayDismiss = this.defaultConfig.overlayDismiss;
+    }
+  }
```
bigopon: we also have this option:
```suggestion
  public overlayDismissChanged(overlayDismiss) {
    if (!overlayDismiss && this.defaultConfig.overlayDismiss) {
      this.overlayDismiss = this.defaultConfig.overlayDismiss;
    }
  }
```
Basically all change handlers are called with signature: `(newValue, oldValue)`. It depends, just in case you are not aware of it.

packages/drawer/src/ux-modal-service.ts
ben-girardet: Can you explain what you mean here ?

The parent is use to `.removeChild()` - to remove the `ux-drawer` on both `dismiss()` and `ok()`

Did I miss something ?

packages/drawer/src/ux-modal-service.ts
ben-girardet: May I explain what is going on here.

When a drawer is dismissed or okeyed (could be inline be being removed from the DOM or from the ModalService), then it will do a few things.

1. It will try to call `canDeactivate` on its attached view-model (if any) (and stop here if returns false or Promise<false> or throw
2. It will call `detached` on its attached view-model (if any)
3. It will call `deactivate` on its attached view-model (if any)
4. It will do all the hiding tasks
5. Once this is completed, the `ux-drawer` dispatch a `CustomEvent` (either `ok` or `dismiss`)
6. Here we find the code above being a listener on the `ux-drawer`. It should be noted that these listeners are only added to `ux-drawer` elements that have been inserted by the ModalService (not inline). Because the ModalService has included the `ux-drawer` element in the DOM, it is now responsible for taking it away with the code below:

```ts
  bindingContext.dismiss = () => {
      drawer.detached();
      const parent = drawer.element.parentNode;
      if (!parent) {
        return;
      }
      parent.removeChild(drawer.element);
      this.eventAggregator.publish(`drawer-${drawerIndex}-resolve`, {
        wasCancelled: true,
        output: null
      });
    }
    bindingContext.ok = (event: CustomEvent) => {
      drawer.detached();
      const parent = drawer.element.parentNode;
      if (!parent) {
        return;
      }
      parent.removeChild(drawer.element);
      this.eventAggregator.publish(`drawer-${drawerIndex}-resolve`, {
        wasCancelled: false,
        output: event.detail
      });
    }
```

I've doing this be getting the `parent` and calling `removeChild`. However I discovered that `detached()` was not called on the `UxDrawer` VM in this particular case. So I thought to add it here manually before doing the actual remove. I did this _before_ the actual removal because I was afraid that doing it after might cause some errors.

Now I am wondering if my approach is correct. Maybe I should remove the `ux-drawer` differently than with a `removeChild()` ? Or should I call `detached()` later on ? Or else ?

packages/drawer/src/ux-modal-service.ts
ben-girardet: I did it with the `drawerIndex` thinking that there only should be one `ModalService` per app (being a singleton). However, I can also inject a transient instance of `EventAggregagor` in this service which then would not be shared with any other parts of the application. Would be safer.

packages/drawer/src/ux-modal-service.ts
ben-girardet: I forgot the `await` in the `layer.drawer.dismiss()` line.

packages/drawer/src/ux-modal-service.ts
ben-girardet: absolutely

packages/drawer/src/ux-drawer.ts
ben-girardet: good point

packages/drawer/src/ux-drawer.ts
ben-girardet: I placed this `+ 10` here as a margin. But to be honest it doesn't quite make sense. I will rather remove it.

packages/drawer/src/ux-drawer.ts
ben-girardet: Haha I tried micro task first, but was too quick and the CSS transition was not running correctly. I will leave a comment

packages/drawer/src/ux-drawer.ts
ben-girardet: I know that changeHandler have the `(newValue, oldValue)` but for some reason I tend not to use it ?? Are there benefits of using the given value instead of the instance value ?

ben-girardet:
Thanks a lot @bigopon for your first round review.

I left a few comments. I am particularly concerned about the discussion on `detached()`. From my tests there is something not going quite right when detaching Drawers created by the ModalService.

For the record:

This is the code that insert the drawer:

```ts
options.host.appendChild(element);
    let childView = this.templatingEngine.enhance({ element: element, bindingContext: bindingContext });
```

Now the question is: what's the right way to do the exact opposite (with full respect of Aurelia Lifecycle).

My current approach (could well be wrong):

```ts
const controllers = (childView as any).controllers as Controller[];
const drawer: UxDrawer =  controllers[0].viewModel as UxDrawer;

drawer.detached();
const parent = drawer.element.parentNode;
if (!parent) {
  return;
}
parent.removeChild(drawer.element);
```

ben-girardet:
@EisenbergEffect  @bigopon 

I think I have some good news. I have [refactored a little the service](https://github.com/aurelia/ux/pull/246/commits/facf984537820d7fec2cc6d618745028e3ce7436) in order to add the `canActivate()` method in the attached VM lifecycle (if any). To be honest I took quite a few parts from the dialog plugin codebase to do so.

This is very useful as a dialog or drawer might not be able to work properly if it doesn't receive the proper params in the `activate()` method. Therefore being able to check this earlier with the standard Aurelia lifecycle method `canActivate()` seems a _non-negotiable_.

The good news that comes with it is that it fixes the [previous bug discovered here](https://github.com/aurelia/ux/pull/246#issuecomment-603161376). So now it seems that anything attached with the service is properly detached and do not interact negatively anymore.

There is however a bad news. And I'm seeking help here. If you try [the demo](https://ux.mintello.com/#/drawer) now and open a crazy drawer (meaning that it opens the current view) you will discover that the chips choices don't work anymore. This is very weird as everything else works fine (checkbox, buttons, ....). I wonder where this could come from. Could it be that the custom attributes (ux-choice-container) are not processed correctly with this new composition method ? Any ideas ?

_As always: I'm open for any feedbacks_

EisenbergEffect:
I did a quick inspection of the DOM and the ux-choice attribute is still present, but with wildly different values, so it appears that something is definitely wrong with that attribute. I wouldn't expect to see it in the DOM at all, since these attributes are removed during compilation (unless I'm getting versions confused in my head).

ben-girardet:
I followed on your findings @EisenbergEffect 

The `ux-choice-container` and `ux-choice` attributes are still present in the DOM for both the working and non-working attributes. I can't tell if they are supposed to stay in the DOM or not for au1 ?

However there is something wrong in the attributes lifecycle when they are part of the dynamic composition. I could identify that for the working attributes (part of the main page) the `attached` and `detached` methods are called. They are not called when the attributes are part of the dynamic composition.

I'm thinking the problem can come from how I set up the `compositionContext`, especially regarding either the `container`, `childContainer` or `bindingContext` value. At the moment as a `container` I pass the `container` value of the `childView` created by the previous `enhance` on the drawer. Not sure if this is correct ?

```ts
  let childView = this.templatingEngine.enhance({ element: element, bindingContext: bindingContext });

    let slot: ViewSlot;
    const controllers = (childView as any).controllers as Controller[];
    const drawer: UxDrawer =  controllers[0].viewModel as UxDrawer;
    try {
      const view: any = controllers[0].view;
      slot = new ViewSlot(view.slots['__au-default-slot-key__'].anchor, false);
    } catch (_error) {
      this.cancelOpening(drawer);
      throw 'Missing slot in drawer';
    }

    let compositionContext = this.createCompositionContext(childView.container as any, element, bindingContext, {
      viewModel: options.viewModel,
      view: options.view,
      model: options.model
    }, slot);
```

```ts
  private createCompositionContext(
    container: any, 
    host: Element, 
    bindingContext: ModalBindingContext, 
    settings: {model?: any, view?: any, viewModel?: any},
    slot? : ViewSlot
    ): CompositionContext {
    return {
      container,
      bindingContext: settings.viewModel ? null : bindingContext,
      viewResources: null as any,
      model: settings.model,
      view: settings.view,
      viewModel: settings.viewModel,
      viewSlot: slot || new ViewSlot(host, true),
      host
    };
  }
```

ben-girardet:
@EisenbergEffect @bigopon 

I would greatly appreciate some help to find out what's wrong with the composition in my code. In order to help debugging I added some `console.log` that you can see [in the demo](https://ux.mintello.com/#/drawer). 

Basically I console.log each attached/detached of `ux-choice-attribute` with an incremental index on each to see when they are properly attaching/detaching or not. Going in and out of the drawer page works fine, also opening an inline drawer, but not with the service using the composition.

I've logged a few steps in the composition process as well.

If you have any ideas for fixing this that would be sooo awesome!!

packages/drawer/src/ux-modal-service.ts
bigopon: We can do:
```ts
    return drawer as UxDrawer & ModalServiceDrawer;
```

packages/drawer/src/ux-drawer.ts
bigopon: generally parameter/local variables minifies better. and also it can avoid overhead of accessing a property. But I think it's not really necessary here. let's ignore this

EisenbergEffect:
@bigopon has more recent experience working inside the templating engine and composition pies. @bigopon Any ideas what might be going on here?

bigopon:
@ben-girardet from your demo, it seems `attached` hasn't been called for the view model of `ux-choice-container`:
image
hence why click events are not registered. Another thing I noted during debugging this issue is that in our components, we are using `click.delegate`. I think for plugin, we should be using `click.trigger` instead, as any code above the component calling `stopPropagation` may potentially affect our handler. Also, in terms of encapsulation, `.trigger` feels better

packages/drawer/src/ux-modal-service.ts
```diff
@@ -178,28 +202,47 @@ export class ModalService {
...
-      bindingContext.composeViewModel.model = options.model;
-    }
 
+    let slot: ViewSlot;
     const controllers = (childView as any).controllers as Controller[];
-    
     const drawer: UxDrawer =  controllers[0].viewModel as UxDrawer;
+    try {
+      const view: any = controllers[0].view;
+      slot = new ViewSlot(view.slots['__au-default-slot-key__'].anchor, false);
```
bigopon:
```suggestion
      slot = new ViewSlot(view.slots[ShadowDOM.defaultSlotKey].anchor, false);
```
We export the slot name here https://github.com/aurelia/templating/blob/5d84c7ee22df8fe92714b8abe99f5ef096542366/src/shadow-dom.js#L435

packages/drawer/src/ux-modal-service.ts
```diff
@@ -178,28 +202,47 @@ export class ModalService {
...
+    } catch (_error) {
+      this.cancelOpening(drawer);
+      throw 'Missing slot in drawer';
```
bigopon: If we need to throw, let's throw a real error object instead
```suggestion
      throw new Error('Missing slot in drawer');
```
otherwise we gonna have some issues with Promise polyfills, and probably in user code too

packages/drawer/src/ux-modal-service.ts
```diff
@@ -169,7 +167,33 @@ export class ModalService {
...
+      viewModel: settings.viewModel,
+      viewSlot: slot || new ViewSlot(host, true),
+      host
+    };
+  }
+
+  private ensureViewModel(compositionContext: CompositionContext): Promise<CompositionContext> {
+    if (typeof compositionContext.viewModel === 'object') {
+      return Promise.resolve(compositionContext);
+    }
```
bigopon: lets experiment with this:
```suggestion
    const viewModel = compositionContext.viewModel;
    if (typeof viewModel === 'object') {
      // emulate an anonymous class that always return the view model object above
      compositionContext.viewModel = function() {
        return viewModel;
      };
    }
```
And then we always pass the composition context through the composition engine

packages/drawer/src/ux-modal-service.ts
```diff
@@ -178,28 +202,47 @@ export class ModalService {
...
+      const canActivate = await invokeLifecycle(compositionContext.viewModel, 'canActivate', options.model);
+      if (!canActivate) {
+        throw new Error('Drawer cannot be opened');
+      }
+    } catch (error) {
+      this.cancelOpening(drawer);
+      throw error;
+    }
+
+    this.compositionEngine.compose(compositionContext).then((controller) => {
```
bigopon: Should this be awaited?

packages/drawer/src/ux-modal-service.ts
bigopon: While I maybe able to guess what's the intention behind creating a slot here and pass to the `createCompositionContext`, it's not obvious and easy to do so. Can you help add some comment here?

packages/drawer/src/ux-modal-service.ts
ben-girardet: I have decided not to `await` here because the resolve of the compose is not required for the `open` to resolve properly. I thought that this could go on asynchronously. I reckon it wouldn't impact much to `await` if there is a reason why it would be better ?

packages/drawer/src/ux-modal-service.ts
ben-girardet: I have tried this experiment. However this does not fix our composition issue yet. It should be noted that the composition issue happened not only for when we pass on a `VM` class but also when we pass a `VM` path or a view only. 

I have updated the [Dumber Gist](https://gist.dumber.app/?gist=938f48392ae45cfeb6d77cef5ab5084c&open=src%2Fapp.ts&open=src%2Fapp.html) to reflect this issue with View Only composition.

packages/drawer/src/ux-modal-service.ts
ben-girardet: Sure. Added comments

ben-girardet:
Thanks @bigopon for your valuable feedbacks. Always good to learn from you!

The composition still fails to properly load the custom attributes :-( 

I have created a [Dumber Gist](https://gist.dumber.app/?gist=938f48392ae45cfeb6d77cef5ab5084c) and a [small git repro](https://github.com/ben-girardet/compo2) for this issue... Maybe easier to spot what's wrong  from one of them..

packages/drawer/src/resources/drawer-util.ts
```diff
@@ -0,0 +1,21 @@
...
+  let element = item;
+  if (element === null) return null;
+  while (element.tagName !== 'BODY' && element.tagName !== 'UX-DRAWER') {
+    element = element.parentElement;
+    if (element === null) return null;
+  }
+  if (element.tagName === 'UX-DRAWER') {
+    const el: any = element;
+    if (
+      el !== null &&
```
bigopon: This is personal preference: I think operator should be at the start of the next line. The reason is we shouldn't have to read the entire line to see the relationship of 2 lines

packages/drawer/src/index.ts
```diff
@@ -0,0 +1,22 @@
+import { DefaultDrawerConfiguration } from './drawer-configuration';
+import { FrameworkConfiguration, PLATFORM } from 'aurelia-framework';
+
+export function configure(frameworkConfig: FrameworkConfiguration, callback?: (config: DefaultDrawerConfiguration) => void) {
+  frameworkConfig.globalResources([
+    PLATFORM.moduleName('./resources/attach-focus-drawer-attribute'),
```
bigopon: Similar to our recent issue with ux-form, we should just use static resources for custom attribute. It's better for optimization, since we already got the decorators on those.

bigopon:
@ben-girardet thanks for the gist, i'll debug it

bigopon:
@ben-girardet The behavior you saw is because `.compose` on composition engine only bind and add a view to a view slot, but does not try to call `attached()` on the controller created. The job of calling `attached` method is delegated to the view slot. So you only need to call `attached()` for our view slot and any views added to it will be invoking `attached` too.

Here's and example: https://gist.dumber.app/?gist=3f3d935b7bc872b8eb59651f8c211620&open=src%2Fapp.ts

packages/drawer/src/index.ts
ben-girardet: Cool. I'll keep this in mind. I'm planning to do some refactoring here to allow for choosing to register this resource or not from plugin config.

ben-girardet:
Thank you so much @bigopon 

This is really helping me!! Looking forward to move forward with this PR.

ben-girardet:
@EisenbergEffect @bigopon 

Here is an important update on the drawer component. 

* I've renamed the component `modal` as it can be used for drawers, but also for dialogs or even menu (new, see below)
* I've added more of the options found in the current `aurelia-dialog` plugin such as a `lock` boolean (when set to false, the modal let the user use the interface in the background)
* It is also possible to set `overlayDismiss: boolean` to decide if the overlay should dismiss the modal or not
* and `outsideDismiss: boolean` to decide if an unlocked modal should be closed when a click is detected outside of itself
* the modal offers now an `openingCallback()` bindable/config that allow the developer to set interact with the modal elements when. This is similar to the `position()` callback of the `aurelia-dialog` plugin

**New: modal as contextual menu**

My vision when starting this drawer component was to create something that works for all kind of situations. I've used the contextual menu situation as a proof of concept. Using the new `openingCallback()` and the `position = 'absolute'` in the modal config it was _pretty easy_ to come up with a function to position the modal next to an anchor.

I've now added a `positionModalWithAnchor()` method in the modal service that do the calculation. Here is the code to open a ViewModel as a menu:

```ts
this.modalService.open({
  viewModel: Hello,
  position: 'absolute',
  openingCallback: (contentWrapperElement: HTMLElement, _overlayElement: HTMLElement) => {
    this.modalService.positionModalWithAnchor(
      anchorElement, 
      contentWrapperElement, 
      {preferedPosition: 'bottom'});
  },
});
```

I would love to hear what you think about all this. Hopefully this is going in a direction that makes sense to you as much as for me :-/ 

**Demo**

As always: https://ux.mintello.com/#/modal (scroll down for the menu demo)

bigopon:
@ben-girardet I haven't reviewed the code, and while playing with the demo, noticed that after opening drawer, menu positioning stops working

ben-girardet:
@bigopon could you give more details on what doesn't work.

Or could it be that when you tried the menu positioning you didn't have the anchors on the screen and it seemed that the positioning didn't work ?

Capture d’écran 2020-04-03 à 07 29 28

ben-girardet:
@bigopon wondering if you still experience the positioning issue ?

bigopon:
I'm getting this issue:
modal_drawer_anchor_err

The first modal trigger button couldn't locate the anchor correctly after open the drawer

ben-girardet:
Hum... are you using Safari @bigopon ? There seem to be something bugging in Safari desktop...

ben-girardet:
The issue that you experience @bigopon comes from a reference issue with the view-model of the drawer. What is happening is 

- the "crazy" drawer is set to open the view-model of the drawer page, meaning that it kind of open its own parent (inception)
- when doing so, it seems that the composition do not create a new instance of the ViewModel but actually uses the already created and used view-model. If I get it well, the view-model instance is now shared between the modal demo page and the modal drawer opened from the "crazy" button
- because anchors are "ref" to the view, they are updated when opening the crazy drawer
- when closing the drawer, these "ref" are lost and positioning doesn't work anymore

Now comes my questions:
- is this normal that a view-model would be shared for several views ? That should not happen right ?
- which would mean that there is something wrong with the composition ? Maybe with the `ensureViewModel()`. Could it be that we are missing a ChildContainer or something like this ?

ben-girardet:
@bigopon nothing wrong with composition, only something wrong with my code :-/

Fixed: https://github.com/aurelia/ux/pull/246/commits/0991be48cb3a05e663ec45f2eb19a3a6ca126159

Do you have any comments regarding the current state of this PR ?

If none, this is my plan:

- make the positioning method better (to include more cases and calculations, like top-left, top-center, top-right)
- create utility to quickly set a modal as either drawer, dialog or menu
- I want to try to create only components like `<ux-drawer>`, `<ux-dialog>` and `<ux-menu>` that are child component of `<ux-modal>`

ben-girardet:
@bigopon @EisenbergEffect 

I've worked on the positioning service. I got inspired by the popper library and in order to avoid dependency I've built something from scratch. I turns out working pretty well. I've already included some uses cases such as:

- position anywhere (12 possibilities)
- offset with the anchor
- margin with the constraint container
- flip placement if overflow with constraint container
- adjust secondary positioning to stay in contraint container if possible
- any constraint container can be chosen (parentElement, other DIV, screen)
- return an instance that can be updated to follow the anchor position (test drag in demo)

It can surely be optimized. But first I wanted to try and have a proof of concept. Open for feedbacks.

Commit: https://github.com/aurelia/ux/pull/246/commits/1b65e4d4e94ad7a2dd64a5a8aed21c5ca662607d

Demo: https://ux.mintello.com/#/anchor

EisenbergEffect:
@ben-girardet Awesome! Thanks for continuing to keep this moving ahead. I seem to have an issue accessing the demo link. Can you confirm that it's correct and working?

bigopon:
@ben-girardet thanks, very nice work. For review purposes, can you help remove all of the dist files? We normally leave it to build process during the release

ben-girardet:
@EisenbergEffect ouch sorry server is down. will fix asap. might take another hour :-/

@bigopon sure, my mistake, normally I leave them out of commits. I've removed them

ben-girardet:
Quick update: I have now finished the ground work for the positioning service and I can use it for positioning modals (not just demo elements).

So I've updated the demo which now use the positioning service on the demo page. Happy to see how it turns out.

https://ux.mintello.com/#/modal

Happy easter everyone !

bigopon:
@ben-girardet very nice. Select modal can also be used as a test subject. And in long term, we would want it that way too

ben-girardet:
@bigopon that would be fantastic! A good real world test!

Would you be willing to team up with me for this? I feel a little lost in the way the select is architectured. 

In order to use the modal the best would be to have a separate VM/V for the options list that should be in the modal. That’s where I need help. 

What would be the steps to make that happen?

The only requirement of the VM modal is to parse the `params` in the `activate` hook to generate its content. So if we can pass on an array of options we should be able to build the view. 

Also the benefit of using a VM/V for this modal is that we could make it configurable so that users could even change the look. Maybe...

bigopon:
@ben-girardet the general theme of `<ux-select/>` is to mimic native `<select/>`, as close, as possible. We will be providing a 2nd component, probably named something close to a "combobox" that has:

- dropdown - select
- text input for filtering
- custom template for option item
- etc...

For the `<ux-select/>` itself, it's mainly this line https://github.com/aurelia/ux/blob/fb4118a38b04c99fe2a74678162d77f1752fbd4f/packages/select/src/ux-select.html#L30 at the moment. You can see that wrapper will be set displayed `fixed` and show/hide accordingly, with some animation. But probably, that's not gonna work properly inside a fixed container. About the container of the option list, there are 2 `<div/>`s, 1 is to position the list relative to anchor, and to show/hide. Another one, inside that, is to take care of the animation.

I'm not sure about how we want to deal with it eventually: (1) only positioning service - extracted & separated from modal, or (2) use the entire modal. For (2), I'd imagine that it'd be a big change in the code, and I'm not sure yet, as atm, we walk the DOM to retrieve stuff I believe. Maybe need to use container to inject stuff now

ben-girardet:
Thanks for your explanation. I also hesitate between 1) and 2). 

1) Using only positioning should be pretty easy. As we don’t have « modal » stuff with it we need to handle the « click outside » to dismiss the wrapper. I think it’s already done btw.

2) Using modal we could provide still a few options to the user such as « lock » or a full screen mode in mobile. As the modal also work inline we can use it to replace the actual wrapper. It would be a bit more work though because then we need to rewrite the open/close/dismiss things.

ben-girardet:
I've given a quick try for the option 1) It's working so so at the moment. 

I feel like I can continue with using only the positioning service but I might need to use the same "trick" than for the modal and move the wrapper as child of the `<body>` before positioning. 

This tend to avoid problems if the select is placed inside a `fixed` container and has the benefit to not alter the `width` and `height` of the `ux-select` element itself (we need correct width and height for the ux-select as it acts as an anchor).

If you think this is too much, I can also wrap the _select_ part inside a `div` and then use this `div` as anchor to have correct width and height. The benefit here is that I don't need to move the `<slot>` container as a child of body.

Option 2) (using modal) would be cool as it already resolves all these issues and create a consistent feel in the UX lib. But here I face an issue with the `<slot>`. 

If I would use the modal component inline, the main idea is to use it with an `if.bind` for opening and closing. It's pretty neat because then it doesn't _polute_ the DOM with elements that are not required when modal is closed. But then the `<slot>` doesn't work because it cannot be placed inside a `if.bind`.

Did you already faced similar situations with `<slot>` used behind a `if.bind` ? Do you have some ideas how to fix this ?

Personally I've already tried something in a component where I placed the `<slot>` in a _fake container_ (hidden). Then I needed to watch this container and ref all its children so that I could move then at the right place when the `if.bind` becomes `true`.

But I feel that this solution is not very stable and would prefer to have something more robust if we go in this direction.

ben-girardet:
@bigopon I've sticked to the modal positioning service and I'm happy with the result. Took me a bit of time to get my head around this select but in facts it's not so complicated...

Please have a look a [the commit](https://github.com/aurelia/ux/pull/246/commits/85372b14d03a838978d3956ed834fe3cd651e194) and [the demo](https://ux.mintello.com/#/components) and tell me what you think.

Note: _I've disabled the collapse on wheel_ now that I can easily update the position of the modal on scroll. It gives a nice effect when you scroll and don't have enough space in the bottom => then the list moves towards the top of the select.

bigopon:
@ben-girardet that looks nice, thanks for doing that. I noticed one difference is that the list modal is positioned at the bottom of the select box, while with material design, I believe it's should be at the start? Also, there's one detail it doesn't have yet: the modal should be opened with selected option align well on the select box. If this feels like off topic of this PR too much, we can defer the select improvement until later. As we need more feature in positioning like offset, maybe?

ben-girardet:
We can change the placement by setting top-start instead of bottom-start when setting the positioning instance. 

As for the precise positioning could you picture what you mean? The positioning service already have the offset setting ready so it might me a matter of setting it correctly.

ben-girardet:
Do you refer to https://material.io/components/menus/ for the specs on select ?

What I understand is that it should have a 0 offset so that the list stick to the select. 

Regarding positioning it all depends on screen real estate around the anchor. We have the flipping feature that does a pretty* good job. I didn't find something saying that it should defaults to top. Seems that defaulting to bottom is fine, as long as there is enough space.

_* The difference I noted from their examples is that when the modal is displayed on top of the select, it should cover the select instead of being placed right on top. However I don't quite like this behavior as it mask some context of the field. But I'm open to add a setting in the positioning service. That would look like setting a `flippingStrategy` config to tell if the flip must cover the anchor or not_

bigopon:
@ben-girardet thanks for the reference link. From there:
image

It seems we can start with the exposed menu style, which is more similar to native select. If folks insists, we can go with dropdown menu style. I originally went with this, but I'm not entirely sure about it being the default behavior

bigopon:
For
> Do you refer to https://material.io/components/menus/ for the specs on select ?

Yes, I originally based the implementation on some other component libs, thinking that is the default, but clearly not necessary.

ben-girardet:
@bigopon the way I read this is:

* Dropdown when opened by an icon or button (or some actions)
* Exposed when linked to a field with the value displayed in it

So for the select we should use exposed scenario. The dropdown will refer to the `menu` component (I hope we could continue the work you started).

bigopon:
@ben-girardet yes let's go with that. Dropdown style for the `<ux-menu/>` sounds valid to me as well 👍

ben-girardet:
I've updated the positioning service with the following

**Missing Space Strategy**
- `flip`: If not enough space in the desired placement, use the flipped side (default)
- `ignore`: don't care about space, place where desired, even if it overflows the constraint element
- `over`: strict respect of the constraint element, event if the positioned element must go over its anchor
- `hide`: hide the element if not enough space (via a customizable class applied to the element)

Note: this strategy is currently only applied to the main placement. It means if we request a `top-start` placement, it will be applied to the `top` but not (yet) the `start` part.

**Demo**
- support touch devices (even though it doesn't display super well on small screens)

https://ux.mintello.com/#/anchor

ben-girardet:
I've implemented the `ignore` strategy for secondary placement. The other ones are not relevant in my opinion.

ben-girardet:
I ping @EisenbergEffect and @bigopon here because I believe the positioning service is **ready for review**. I have some questions below, but first, in a nutshell, here is what this service does:

* Creates a positioning instance from an anchor and an element
* Will position the element close to the anchor according to the chosen options
* The element will be correctly positioned on first instance creation and won't be moved on scroll or resize, this is left up to the developer(*)
* The instance has a public `update()` api that re-positioned the element according to the options
* Each option is exposed as public on the instance so that developer can change it at any time

(*) The reasons why I chose (for now) not to handle scroll or resize is that I prefer to leave the positioning focused on what it's supposed to achieve. Listening to such events and update automatically could be nice but in order to cover each edge case it could end up being much work.

**Positioning Options**

* `placement?: UxModalPlacement`: the preferred placement for the element, see below (default: `auto`)
* `offsetX?: number`: horizontal offset from the anchor in pixels (default: `5`)
* `offsetY?: number`: vertical offset from the anchor in pixels (default: `5`)
* `constraintElement?: HTMLElement | Window;`: which element should limit the positioning of the element (default: `element.parentElement`)
* `constraintMarginX?: number`: horizontal margin in pixels towards the constraint before we assume there is not enough space  (default: `5`)
* `constraintMarginY?: number`: vertical margin in pixels towards the constraint before we assume there is not enough space  (default: `5`)
* `missingSpaceStrategy?: 'flip' | 'ignore' | 'hide' | 'over'`: strategy to handle what happen when there is not enough space between the anchor and the constraint element (default: `flip`)
* `hiddenClass?: string`: CSS class added to the element when it should be hidden (default: `ux-positioning--hidden`)

`UxModalPlacement`
* `auto`
* `auto-start`
* `auto-end`
* `left`
* `left-start`
* `left-end`
* `right`
* `right-start`
* `right-end`
* `top`
* `top-start`
* `top-end`
* `bottom`
* `bottom-start`
* `bottom-end`

**A few questions for the review**

1. **Package**: As this service is autonomous, I don't think it should stay in the `modal` package. I am tempted to create a `positioning` package inside the learn mono-repo and have this service stand alone. The benefit is that this service could be used without any of the rest of the library if this is what someone wants. Also, it will be useful in other packages, such as the `select` component and the rest of the `modal` component is not required. 
2. **Naming**: I've noticed that Aurelia doesn't have much classes called `....Service` but rather `...Engine`. Maybe this would be better called `UxPositioningEngine`
3. **Factory**: In order to work well with the DI (because I need to inject `TaskQueue`), I've created a single class that have a `getInstance()` public method to create an instance correctly linked to the DI. But I believe there are better patterns as Factory for this kind of things. Do you have any advice to make this better ?

bigopon:
For naming, probably @EisenbergEffect can tell, I'm leaning towards engine as well. We do have `DialogService`, and this is gonna probably be `ModalService` as well, but `PositioningEngine` sounds closer to `StyleEngine`

For packaging, I'm ok with having a separate package for it, though it's not my call 😁 @EisenbergEffect . Maybe we can invite some people to try it out in their non-Aurelia projects and help spot bugs.

What is the last point about?

EisenbergEffect:
Pulling it out into its own package seems fine. Not sure about "engine" or "service" actually. You could just call it "ElementPositioner".

packages/modal/src/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  frameworkConfig.globalResources([
+    AttachFocusAttribute,
+    DismissModalAttribute,
+    OkModalAttribute, 
+    PLATFORM.moduleName('./ux-modal'),
+  ]);
+  const config = frameworkConfig.container.get(UxDefaultModalConfiguration);
+  if (typeof callback === 'function') {
+    callback(config);
+  }
```
bigopon, line 17:
```suggestion
  if (typeof callback === 'function') {
    const config = frameworkConfig.container.get(UxDefaultModalConfiguration);
    callback(config);
  }
```

packages/modal/src/positioning/ux-modal-positioning.ts
```diff
@@ -0,0 +1,409 @@
...
+  private constraintHeight: number = 0;
+  private constraintWidth: number = window.innerWidth;
+  private spaceTop?: number = void 0;
+  private spaceRight?: number = void 0;
+  private spaceBottom?: number = void 0;
+  private spaceLeft?: number = void 0;
+
+  constructor(public taskQueue: TaskQueue) {}
+
+  public getInstance(anchor: HTMLElement, element: HTMLElement, options?: UxPositioningOptions) {
```
bigopon, line 48: Should we make this method `static`, with an extra parameter for `TaskQueue` instead?

packages/modal/src/positioning/ux-modal-positioning.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  private resetElement() {
+    this.element.style.position = 'absolute';
+    this.element.style.top = '0';
+    this.element.style.left = '0';
+    this.element.style.right = 'auto';
+    this.element.style.bottom = 'auto';
+    this.element.style.width = 'unset';
+    this.element.style.height = 'unset';
```
bigopon, line 137: I'm not sure about this, but it feels a bit repetitive. Maybe:
```suggestion
    const style =  this.element.style;
    style.position = 'absolute';
    style.top = '0';
    left = '0';
    right = 'auto';
    bottom = 'auto';
    width = 'unset';
    height = 'unset';
```
Better for minification/perf at runtime, since less property accesses calls. Probably irrelevant to app perf, but pursuing "every bit counts" religiously here won't hurt? I'm not insisting on this though.

packages/modal/src/positioning/ux-modal-positioning.ts
```diff
@@ -0,0 +1,409 @@
...
+    return this.spaceTop;
+  }
+  
+  private getSpaceRight() {
+    if (this.spaceRight !== undefined) {
+      return this.spaceRight;
+    }
+    this.spaceRight = this.constraintElement === window ?
+      window.innerWidth - window.scrollX - this.anchorRect.width - this.anchorRect.left - this.constraintX - this.constraintMarginX :
+      this.constraintX + this.constraintWidth - this.anchorRect.width - this.anchorRect.left - this.constraintMarginX;
```
bigopon, line 206: Let's apply the same logic with coding style to ternary: relation ship of lines should be easy to skim, for that, let's put the `?` and `:` in the front:
```suggestion
    this.spaceRight = this.constraintElement === window
      ? window.innerWidth - window.scrollX - this.anchorRect.width - this.anchorRect.left - this.constraintX - this.constraintMarginX
      : this.constraintX + this.constraintWidth - this.anchorRect.width - this.anchorRect.left - this.constraintMarginX;
```

packages/modal/src/ux-modal-service.ts
```diff
@@ -0,0 +1,493 @@
...
+    }
+    return undefined;
+  }
+
+  public getLastLayer(): UxModalLayer {
+    return this.modalLayers[this.modalLayers.length - 1];
+  }
+
+  public getLastModal(): UxModal {
+    return this.getLastLayer().modal;
```
bigopon, line 173: If this method is not guaranteed to be called with at least 1 layer, then this line could break

packages/modal/src/ux-modal.ts
```diff
@@ -0,0 +1,344 @@
...
+      if (this.defaultConfig.position !== undefined) {
+        this.position = this.defaultConfig.position
+      }
+      if (this.defaultConfig.keyboard !== undefined) {
+        this.keyboard = this.defaultConfig.keyboard
+      }
+      if (this.defaultConfig.theme !== undefined) {
+        this.theme = this.defaultConfig.theme
+      }
+    }
```
bigopon, line 78: nit: this `}` seems off

packages/modal/src/ux-modal.ts
```diff
@@ -0,0 +1,344 @@
...
+    this.styleEngine.applyTheme(newValue, this.element);
+  }
+
+  public setViewportType() {
+    this.viewportType = window.innerWidth < this.modalBreakpoint ? 'mobile' : 'desktop';
+  }
+
+  @computedFrom('type', 'viewportType')
+  public get computedType(): 'standard' | 'modal' {
+    return this.viewportType === 'mobile' ? 'modal' : this.type;
```
bigopon, line 270: Do we need to normalize `this.type` return here to either `standard` or `modal`?

ben-girardet:
Wow thanks for this great review ! I'll look in detail at each comments.

packages/modal/src/positioning/ux-modal-positioning.ts
ben-girardet: My question (3) about factoring was linked to your question here. I would prefer to have a static `getInstance` or `create` method. In order to go in this direction I suggest to drop the TaskQueue dependency and replace the only usage by a `setTimeout(/*..*/, 0)` which has the same effect (tested).

This would leave us with a service that has no dependency and that can be used in any application (aurelia or not)

What do you think  @bigopon  ?

app/src/modal-comp.ts
```diff
@@ -0,0 +1,149 @@
+import { UxModalService, UxModalPosition, UxModalPositioning, UxModalPlacement, UxPositioningOptions } from '@aurelia-ux/modal';
+import { Hello } from './hello';
+import { inject } from 'aurelia-framework';
+
+@inject(UxModalService, UxModalPositioning)
+export class ModalComp {
+
+  public host: '' | '.modal-host' = '';
```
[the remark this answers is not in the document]
ben-girardet: Absolutely no valid reason :-/

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+  private showing: boolean = false;
+  private hiding: boolean = false;
+
+  constructor(
+    public element: HTMLElement,
+    private styleEngine: StyleEngine,
+    private modalService: ModalService,
+    private taskQueue: TaskQueue,
+    private defaultConfig: DefaultDrawerConfiguration) {
+      if (this.defaultConfig.modalBreakpoint !== undefined) {
```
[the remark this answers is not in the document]
ben-girardet: that's right !

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+      if (this.defaultConfig.keyboard !== undefined) {
+        this.keyboard = this.defaultConfig.keyboard
+      }
+      if (this.defaultConfig.theme !== undefined) {
+        this.theme = this.defaultConfig.theme
+      }
+    }
+  private bindingContext: any;
+  public bind(bindingContext: any) {
+    this.bindingContext = bindingContext;
```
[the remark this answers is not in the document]
ben-girardet: In this particular case I'm not quite sure it's required as the `bindingContext` is use here for internal purposes but not in order to expose it for consumption. But I'm not expert with binding and the `overrideContext` concept.

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  private moveToHost() {
+    const host = this.getHost();
+    if (!host) {
+      return
+    };
+    host.appendChild(this.element);
+  }
+
+  private removeFromHost() {
```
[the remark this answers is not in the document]
ben-girardet: In a previous attempt at such a component I did need to bring back the element to its original position. However now with the concept that the drawer only exists in the DOM if it is displayed then it seems not required. As you said, we can experiment and keep an eye one it. I'll leave a comment in the code.

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+      return
+    };
+    try {
+      host.removeChild(this.element);
+    } catch (e) {
+      // if error, it's because the child is already removed
+    }
+  }
+
+  private getHost(): Element | undefined {
```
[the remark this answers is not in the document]
ben-girardet: Will be commit from my code change

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+    if (this.host === 'body') {
+      return document.body;
+    } else if (this.host instanceof HTMLElement) {
+      return this.host;
```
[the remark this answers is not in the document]
ben-girardet: I don't think that works. We can't return `body` if the signature is `Element | null`

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+    this.styleEngine.applyTheme(newValue, this.element);
+  }
+
+  public setViewportType() {
+    this.viewportType = window.innerWidth < this.modalBreakpoint ? 'mobile' : 'desktop';
+  }
+
+  @computedFrom('type', 'viewportType')
+  public get computedType(): 'standard' | 'modal' {
+    return this.viewportType === 'mobile' ? 'modal' : this.type;
```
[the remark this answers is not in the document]
ben-girardet: It works without but it's way better and cleaner to do so. Thanks!

packages/drawer/src/ux-drawer.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  public stop(event: Event) {
+    event.stopPropagation();
+  }
+
+  private getAnimationDuration() {
+    const overlayElementDuration: string = window.getComputedStyle(this.overlayElement).transitionDuration || '';
+    const contentDuration: string = window.getComputedStyle(this.contentElement).transitionDuration || '';
+    // overlayElementDuration and contentDuration are string like '0.25s'
+    return Math.max(parseFloat(overlayElementDuration), parseFloat(contentDuration)) * 1000;
```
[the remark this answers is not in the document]
ben-girardet: I left the following comment in the code:

```ts
// In order to allow precise animation we allow different duration
// value for animating the overlay and the drawer. In most cases it will
// be the same value but we can imagine a fast overlay and slower modal
// apearence for exemple
// Because the duration is used to determine when we can safely assume the
// modal disappeard, we only keep the maximum value.
```

packages/modal/src/positioning/ux-modal-positioning.ts
ben-girardet: I have applied this same logic to other places where `this.element.style` is used. I'm event tempted to ref this property as a private `elementStyle` instance property. I did not do it yet.

packages/modal/src/ux-modal-service.ts
ben-girardet: good catch. fixed

ben-girardet:
Thanks again @bigopon for all your valuable comments. I've replied / fixed them. 

I was not clear, but the current drawer package is out of date as everything is currently in the `modal` package. But don't worry, I have applied all your comments directly to the `modal` package.

From now on, in order to bring a little more clarity on this PR I suggest to create two distinct PRs and close this one as WIP without merging.

**1. Positioning PR**
With a new `positioning` package containing only the positioning service. 

I tend to prefer the `PositioningService` or `PositioningEngine` names which seem to follow a little more the other naming in Aurelia codebase. But of course I leave the last call to @EisenbergEffect

**2. Modal PR**
With the `modal` package containing the current implementation of the component and service. I'm planning to finish up a few more things before final review.

Would you agree with this course of action ? If so we could probably merge the positioning PR pretty soon.

packages/modal/src/positioning/ux-modal-positioning.ts
bigopon: Thanks, I see what you meant with 3 now. I was vaguely guessing that as well. For that, I think we can use an interface here instead. Instead of expecting a `TaskQueue`, we can expect an object with following shape:
```ts
interface ITaskQueue {
  queueTask(task: () => void): void;
}
```
maybe that'd work as a first step? We don't have to solve all the reuse-ability issues for now.

bigopon: Also, folks still have Popper if they want something.

ben-girardet: If we go static I don't think it's a good idea to expect anything like a TaskQueue. As a consumer of this service I would not want to always have to pass on a taskqueue instance. This should be handled by the service.

We could however add an option in the `options` parameter where one could pass on the taskQueue if wanted, otherwise it defaults to `setTimeout(..., 0)`.

Or we can keep it non-static and have a stronger dependency with Aurelia codebase.

Also, the taskQueue is in fact only needed for a very specific use case which do not occur all the time and only happens once during the init phase. So placing here a timeout 0 doesn't hurt much ?!?

bigopon: Ok lets go with a timeout for now and add TQ later. The issue I have with it is its current way of init: create a empty instance, and then passing the injected task queue around. It feels patchy/hacky to me. Maybe lets ping @EisenbergEffect if he has any input around the API

ben-girardet: Yes, it is patchy/hacky and that was where I was looking for a good factory pattern. If we go static then we don't  need anything more. Otherwise we might need two classes, first a factory service that create a proper instance of the actual positioning class.

bigopon:
@ben-girardet this is looking good, but your new plan sounds better to me. Scrolling perf got degrade d a bit in this big PR :D

packages/modal/src/positioning/ux-modal-positioning.ts
EisenbergEffect: If you use the factory resolver, it can give you a function that will create an instance of UxModalPositioning, allowing you to supply the anchor, element, and options.

See https://github.com/aurelia/dependency-injection/blob/master/src/resolvers.ts#L341

ben-girardet: Thanks @EisenbergEffect for pointing to the `Factory` resolver.

ben-girardet:
@bigopon 

Positioning PR ready here: https://github.com/aurelia/ux/pull/254

If it's merged rapidly I will wait on the merge before to set the new Modal PR (to avoid copying the positioning package in the modal branch if possible). If we need time before merge I will copy anyway and work on the other PR.

ben-girardet:
@bigopon thanks for the merge of #254 

Modal PR is now available here: #255 

_I will close this PR as now it is replaced by the two mentioned above._

PR #249, opened 2020-04-02
Fix date

ben-girardet:
Because I'm using the `datepicker` in localized website I came to realize a pretty annoying bug. 

Basically the `parsers` and `formatters` bindable are conflicting and create weir behavior. For exemple, if I set the `formatters.bind="{date: 'DD-MM-YYYY'}"` it all works when until the field is blurred. Then the property set value from the formatter is replace by a value parsed again but withe the `parsers` value.

Hopefully it makes sense and this fix can be released soon.

I also added a fix to the theme where the picker needs a background property not connected to the default controls. It's better to keep this picker background in sync with other picker theme properties and have nice defaults that work well.

### DIFF
--- a/packages/datepicker/src/ux-datepicker.ts
+++ b/packages/datepicker/src/ux-datepicker.ts
@@ -45,10 +45,6 @@ export class UxDatepicker implements UxInputComponent {
     datetime: 'L LT'
   };
 
-  @bindable public parsers = {
-    time: ['h:m a', 'H:m']
-  };
-
   @bindable({ defaultBindingMode: bindingMode.twoWay })
   public value: any;
 
@@ -92,13 +88,13 @@ export class UxDatepicker implements UxInputComponent {
     }
 
     if (this.minTime != null) {
-      const dateParse = moment(this.minTime, this.parsers.time);
+      const dateParse = moment(this.minTime, this.formatters.time);
 
       this.minTime = dateParse.isValid() ? dateParse : null;
     }
 
     if (this.maxTime != null) {
-      const dateParse = moment(this.maxTime, this.parsers.time);
+      const dateParse = moment(this.maxTime, this.formatters.time);
 
       this.maxTime = dateParse.isValid() ? dateParse : null;
     }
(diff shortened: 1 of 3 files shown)


ben-girardet:
@bigopon @EisenbergEffect 

Ready for review

packages/datepicker/src/ux-datepicker.ts
```diff
@@ -146,7 +142,13 @@ export class UxDatepicker implements UxInputComponent {
...
     let parseValue;
 
-    parseValue = this.type === 'time' ? moment(this.textboxValue, this.parsers.time) : moment(this.textboxValue);
+    if (this.type.toLowerCase() === 'date') {
+      parseValue = moment(this.textboxValue, this.formatters.date);
+    } else if (this.type.toLowerCase() === 'time') {
+      parseValue = moment(this.textboxValue, this.formatters.time);
+    } else {
+      parseValue = moment(this.textboxValue, this.formatters.datetime);
+    }
```
bigopon, line 151:
```suggestion
    const pickerType = this.type.toLowerCase();
    if (pickerType === 'date') {
      parseValue = moment(this.textboxValue, this.formatters.date);
    } else if (pickerType === 'time') {
      parseValue = moment(this.textboxValue, this.formatters.time);
    } else {
      parseValue = moment(this.textboxValue, this.formatters.datetime);
    }
```

bigopon: perf difference will be minor, though we don't want to repeatedly call `.toLowerCase()`

EisenbergEffect:
I agree. I think the code looks fine, but I'd want @serifine to confirm that we aren't missing something by removing the parsers.

packages/datepicker/src/ux-datepicker.ts
ben-girardet: Yes this can be optimized. I've done it by calling `.toLowerCase()` on `typeChanged()` so that it's really only called once per type change and not for all valueChanged ...

ben-girardet:
`parsers` is not anywhere else in the date picker code. Probably was left in for legacy purposes. 

Let's wait for @serifine answer about this 👍

serifine:
I'm not entirely sure why I put that in to be honest. Try removing it and
see if it still works as intended.

bigopon:
Looks good to me then. Thanks @serifine

PR #250, opened 2020-04-02
fix(choice): process value change is new !== old

ben-girardet:
Here is a fix for `ux-choice-container` attribute. When we optimized the code in `valueChanged()` we removed an important part... this valueChanged() must also request to process the value if the `newValue` is different thant `oldValue`.

Currently the attribute works well for one-way binding because the right choice is selected anyway. But for two-way binding the valueChanged() needs to process the value.

Ready for review @EisenbergEffect @bigopon

### DIFF
--- a/packages/core/src/components/ux-choice-container-attribute.ts
+++ b/packages/core/src/components/ux-choice-container-attribute.ts
@@ -82,7 +82,7 @@ export class UxChoiceContainerAttribute {
     }
   }
 
-  public valueChanged(newValue: string | string []) {
+  public valueChanged(newValue: string | string[], oldValue?: string | string[]) {
     if (this.multiple === 'auto') {
       this.multipleChanged(); // call this to ensure isMultiple respect value type
     }
@@ -91,6 +91,8 @@ export class UxChoiceContainerAttribute {
       this.requestProcessValue();
     } else if (!this.isMultiple && Array.isArray(newValue)) {
       this.value = undefined;
+      this.requestProcessValue();
+    } else if (!oldValue || newValue !== oldValue) {
       this.requestProcessValue();
     }
   }
--- a/packages/core/src/components/ux-choice-container-attribute.ts
+++ b/packages/core/src/components/ux-choice-container-attribute.ts
@@ -82,19 +82,16 @@ export class UxChoiceContainerAttribute {
     }
   }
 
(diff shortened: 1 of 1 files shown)


bigopon:
By default, `valueChanged` wont be called unless `newValue !== oldValue`. Can you give an example where this would not be the case?

ben-girardet:
I don't think it would happen. We can remove this `if`

packages/core/src/components/ux-choice-container-attribute.ts
```diff
@@ -82,19 +82,16 @@ export class UxChoiceContainerAttribute {
...
       this.value = [];
-      this.requestProcessValue();
     } else if (!this.isMultiple && Array.isArray(newValue)) {
       this.value = undefined;
-      this.requestProcessValue();
-    } else if (!oldValue || newValue !== oldValue) {
-      this.requestProcessValue();
     }
+    this.requestProcessValue();
   }
```
bigopon: Im a bit confused by the logic inside this method. Can you help give some comments, with regards to the logic how we update and request change? One of the confusion I have is how we updating `value` inside a value change, and also immediately requesting process value.

ben-girardet: I've updated the PR with comments.

Also because the `requestProcessValue()` is behind a `TaskQueue` then it will only be _processed_ once the value is really set properly.

ben-girardet: Oh but I see. You probably mean that we could `return` after changing the `value` instead of letting go the method to the end and request the process.... I guess that would be better ?

ben-girardet: `return` when changing value in valueChanged (https://github.com/aurelia/ux/pull/250/commits/4344363cb65d3d4fd7880c08c00f4be8feca5537)

bigopon: Yes, that is what I meant. Can you try and and see if it works better?

ben-girardet: Yes I tried. It works fine. Not better as both work but with the `return` it will have slightly better perf.

bigopon:
kk thanks for the fix @ben-girardet 
I will try to setup test infra for us to start testing and be able to solidify our logic around tricky parts soon.

PR #252, opened 2020-04-14
fix(chip-input): stop propagation on ENTER or ESC

ben-girardet:
I noticed that when using a `chip-input` inside a drawer we get a undesired behavior. Both are listening for the keyup change on ENTER and ESC keys.

I suggest that when the chip-input detects the ENTER or ESC keyup it stops the propagation so that other handlers don't act on it as well.

@bigopon what do you think ?

### DIFF
--- a/packages/chip-input/src/ux-chip-input.ts
+++ b/packages/chip-input/src/ux-chip-input.ts
@@ -80,6 +80,7 @@ export class UxChipInput implements UxInputComponent {
 
     if (key === 13) {
       this.addChip();
+      event.stopImmediatePropagation();
     }
 
     if (key === 37) {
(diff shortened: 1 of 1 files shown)


bigopon:
This sounds good to me. Does `stopImmediatePropagation` work, should it be `stopPropagation`?

ben-girardet:
From my understanding both work, `stopImmediatePropagation` is a little more aggressive in terms of stopping the event.

https://stackoverflow.com/questions/5299740/stoppropagation-vs-stopimmediatepropagation

bigopon:
Sounds good then. I misunderstood the `stopImmediatePropagation`, thought it was only for the current target.

bigopon:
@EisenbergEffect pinging you to queue a release

bigopon:
Thanks @ben-girardet for this PR. I think generally we can stop the (immediate)propagation for the enter key, and probably a few more if needed. It's suitable action for a component lib.

PR #255, opened 2020-04-20
Modal component and service

ben-girardet:
Here is the PR focused on the modal component and service. I believe that it is ready for final review. It might continue to evolve but I think it's at a state that is robust and useable.

Once merged I would like to use it to finish up the work on the [Demo PR](https://github.com/aurelia/ux/pull/241).

### DIFF
--- a/packages/modal/src/index.ts
+++ b/packages/modal/src/index.ts
@@ -1,0 +1,24 @@
+import { OkModalAttribute } from './resources/ok-modal-attribute';
+import { DismissModalAttribute } from './resources/dismiss-modal-attribute';
+import { AttachFocusAttribute } from './resources/attach-focus-attribute';
+import { UxDefaultModalConfiguration } from './ux-modal-configuration';
+import { FrameworkConfiguration, PLATFORM } from 'aurelia-framework';
+
+export function configure(frameworkConfig: FrameworkConfiguration, callback?: (config: UxDefaultModalConfiguration) => void) {
+  frameworkConfig.globalResources([
+    AttachFocusAttribute,
+    DismissModalAttribute,
+    OkModalAttribute,
+    PLATFORM.moduleName('./ux-modal'),
+  ]);
+  if (typeof callback === 'function') {
+    const config = frameworkConfig.container.get(UxDefaultModalConfiguration);
+    callback(config);
+  }
+
+}
+
+export { UxModalTheme } from './ux-modal-theme';
+export { UxModal } from './ux-modal';
+export * from './ux-modal-service';
+export * from './ux-modal-configuration';
--- a/packages/modal/src/resources/attach-focus-attribute.ts
+++ b/packages/modal/src/resources/attach-focus-attribute.ts
@@ -1,0 +1,16 @@
+import { customAttribute, inject } from "aurelia-framework";
+
+@inject(Element)
+@customAttribute('attach-focus')
+export class AttachFocusAttribute {
+
+  public value: boolean | string;
+
+  constructor(private element: HTMLElement) {}
+
+  public attached() {
+    if (this.value === '' || this.value === 'attach-focus' || (this.value && this.value !== 'false')) {
+      this.element.focus();
+    }
+  }
+}
--- a/packages/modal/src/resources/attach-focus-attribute.ts
+++ b/packages/modal/src/resources/attach-focus-attribute.ts
@@ -1,7 +1,7 @@
 import { customAttribute, inject } from "aurelia-framework";
 
 @inject(Element)
-@customAttribute('attach-focus')
+@customAttribute('ux-attach-focus')
 export class AttachFocusAttribute {
 
   public value: boolean | string;
--- a/packages/modal/src/resources/attach-focus-attribute.ts
+++ b/packages/modal/src/resources/attach-focus-attribute.ts
@@ -9,7 +9,7 @@ export class AttachFocusAttribute {
   constructor(private element: HTMLElement) {}
 
   public attached() {
-    if (this.value === '' || this.value === 'attach-focus' || (this.value && this.value !== 'false')) {
+    if (this.value === '' || (this.value && this.value !== 'false')) {
       this.element.focus();
     }
   }
--- a/packages/modal/src/resources/modal-util.ts
+++ b/packages/modal/src/resources/modal-util.ts
@@ -1,0 +1,28 @@
+import { UxModal } from '../ux-modal';
+
+export function findModal(item: HTMLElement | null): UxModal | null {
+  let element = item;
+  if (element === null) return null;
+  while (element.tagName !== 'BODY' && element.tagName !== 'UX-DRAWER') {
+    element = element.parentElement;
+    const isModaleElement = element ? modalViewModelFromElement(element) : false;
+    if (isModaleElement !== false) {
+      return isModaleElement;
+    }
+    if (element === null) return null;
+  }
+  return null;
+}
+
+export function modalViewModelFromElement(element: HTMLElement): UxModal | false {
+  const el: any = element;
+  if (!el.au) {
+    return false;
+  }
+  for (let key in el.au) {
+    if (el.au[key].viewModel && el.au[key].viewModel instanceof UxModal) {
+      return el.au[key].viewModel
+    }
+  }
+  return false;
+}
--- a/packages/modal/src/resources/modal-util.ts
+++ b/packages/modal/src/resources/modal-util.ts
@@ -5,8 +5,8 @@ export function findModal(item: HTMLElement | null): UxModal | null {
   if (element === null) return null;
   while (element.tagName !== 'BODY' && element.tagName !== 'UX-DRAWER') {
     element = element.parentElement;
-    const isModaleElement = element ? modalViewModelFromElement(element) : false;
-    if (isModaleElement !== false) {
+    const isModaleElement = element ? modalViewModelFromElement(element) : null;
+    if (isModaleElement !== null) {
       return isModaleElement;
     }
     if (element === null) return null;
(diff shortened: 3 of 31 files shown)


packages/modal/src/index.ts
```diff
@@ -0,0 +1,24 @@
+import { OkModalAttribute } from './resources/ok-modal-attribute';
+import { DismissModalAttribute } from './resources/dismiss-modal-attribute';
+import { AttachFocusAttribute } from './resources/attach-focus-attribute';
+import { UxDefaultModalConfiguration } from './ux-modal-configuration';
+import { FrameworkConfiguration, PLATFORM } from 'aurelia-framework';
+
+export function configure(frameworkConfig: FrameworkConfiguration, callback?: (config: UxDefaultModalConfiguration) => void) {
+  frameworkConfig.globalResources([
+    AttachFocusAttribute,
```
bigopon, line 9: This will collide with anyone who uses `aurelia-dialog` together with this, is this suggesting that the modal service can cover all use cases of dialog?

ben-girardet: Good question. I believe it's very close but still I prefer to ensure it doesn't collide. First I wanted to name it differently (by prefixing something like `ux-modal-...` but then I realized that this attribute is, in fact, not so related to `aurelia-dialog` or `ux-modal` as it only call the element `focus()` api.

So my plan was to make it optional and let user opt-in/out for this attribute (as aurelia-dialog does as well). I can do it with a `useDefault()` pattern in the config class. Would that be ok ?

Thanks for catching this !

bigopon: I see, so maybe we can do this:
```ts
frameworkConfig.postTask(() => {
  const globalResources = frameworkConfig.container.get(ViewResources);
  if (globalResources.getAttribute('attach-focus') === null) {
    globalResources.registerAttribute(
      'attach-focus',
      metadata.getOwn(metadata.resource, AttachFocusAttribute)
    );
  }
})
```

bigopon: Sorry, i just realize this probably wont work, since the views will be compiled before the attribute is ready

ben-girardet: Ohh.. this is brilliant. Didn't know about `postTask()`. I tried to use the `registerAttribute()` method but failed to really understand it. I'm missing a third parameter to the method and I don't know how to build the `metadata` object ?

What if I do: 

```ts
  frameworkConfig.postTask(() => {
    const globalResources = frameworkConfig.aurelia.resources;
    if (globalResources.getAttribute('attach-focus') === null) {
      const extraConfig = new FrameworkConfiguration(frameworkConfig.aurelia);
      extraConfig.globalResources(AttachFocusAttribute);
      extraConfig.apply();
    }
  });

```

ben-girardet: Oh ok.. shame I liked it !

bigopon: `metadata` is an import from `aurelia-metadata`. For the 3rd parameter, I think we don't need it. Probably we will need to prefix it. Not the best, but consistently good I think

ben-girardet:
Regarding Modal VS Dialog comparison. I can point to two things that are different (and might stay different).

1. Dialog has some variants in how we can call the `open()` method. 

Depending on how it is called it will return and handle the returned promise a little differently. I am not a big user of Aurelia Dialog so I don't know how important these variants are. In my opinion I prefer when a service like this has one consistent API. Therefore I would rather stick with a single `open()` method that works in the same manner all the time.

2. The Dialog attaches the `whenClosed()` method to the returned `Promise<>` from the `open()` api. This comes handy as it requires one less `then()` or `await` but I prefer the way the `modalService` return a standard Promise pointing to the `UxModal` ViewModel. When consumed, the `open()` api requires one more `then()` or `await` but then the returned value can do more than just `whenClosed()` => it gives full access to the ViewModel.

Both of these could lead to breaking changes when switching from Aurelia Dialog to Aurelia Ux Modal but they should not remove any important functionality that I am aware of.

packages/modal/src/index.ts
ben-girardet: ok, I'll go for the prefix version

packages/modal/src/resources/attach-focus-attribute.ts
```diff
@@ -0,0 +1,16 @@
...
+@inject(Element)
+@customAttribute('ux-attach-focus')
+export class AttachFocusAttribute {
+
+  public value: boolean | string;
+
+  constructor(private element: HTMLElement) {}
+
+  public attached() {
+    if (this.value === '' || this.value === 'attach-focus' || (this.value && this.value !== 'false')) {
```
bigopon, line 12: This value `attach-focus` here is not needed. 
```suggestion
    if (this.value === '' || (this.value && this.value !== 'false')) {
```

packages/modal/src/resources/modal-util.ts
```diff
@@ -0,0 +1,28 @@
...
+    const isModaleElement = element ? modalViewModelFromElement(element) : false;
+    if (isModaleElement !== false) {
+      return isModaleElement;
+    }
+    if (element === null) return null;
+  }
+  return null;
+}
+
+export function modalViewModelFromElement(element: HTMLElement): UxModal | false {
```
bigopon, line 17: an `object | false` return type is quite a weird API to have. I think we can return `null` instead, and in the other check, it can be done against `null`, should be easier to guess, consistent with anywhere else, either in/out of Aurelia

packages/modal/src/ux-modal-service.ts
```diff
@@ -0,0 +1,402 @@
...
+
+  private createModalElement(options: UxModalServiceOptions, bindingContext: UxModalBindingContext): HTMLElement {
+    const element = document.createElement('ux-modal');
+    element.setAttribute('dismiss.trigger', 'dismiss()');
+    element.setAttribute('ok.trigger', 'ok($event)');
+    if (options.position !== undefined) {
+      element.setAttribute('position', options.position);
+    }
+    if (options.moveToBodyTag === false) {
+      element.setAttribute('move-to-body-tag.bind', 'false');
```
bigopon, line 191: - do we need the `-tag` suffix?
- I couldn't find this `move-to-body-tag` in `ux-model.ts`, is this needed? What is the default behavior for this?

ben-girardet: This whole `move-to-body-tag` is not required anymore. 

I had renamed the method `moveToHost()` and forgot to change here. But then I also removed the option to decide to move it or not. In fact it's handled by the `host` property => if set, we move the modal to its host without choice.

Thanks for catching this !

packages/modal/src/resources/modal-util.ts
```diff
@@ -0,0 +1,28 @@
+import { UxModal } from '../ux-modal';
+
+export function findModal(item: HTMLElement | null): UxModal | null {
+  let element = item;
+  if (element === null) return null;
+  while (element.tagName !== 'BODY' && element.tagName !== 'UX-DRAWER') {
```
bigopon, line 6: Can we have some comments describing what would happen when we hit a `ux-drawer` here? So the reason why the check against `ux-drawer` can be easier to understand

packages/modal/src/ux-modal-service.ts
```diff
@@ -0,0 +1,403 @@
...
+    if (index !== -1) {
+      this.modalLayers.splice(index, 1);
+    }
+    if (this.modalLayers.length === 0) {
+      this.removeListener();
+    }
+  }
+
+  private setListener() {
+    window.addEventListener('keyup', this);
```
bigopon, line 134: why is key on window and click on document? If we do both of them in either window or document, I think we can use `this` as listener for both and switch `event.type` to call the corresponding handler instead

packages/modal/src/ux-modal-service.ts
```diff
@@ -0,0 +1,403 @@
...
+    container: any, // there is a TS error if this is not any ?
+    host: Element,
+    bindingContext: UxModalBindingContext,
+    settings: {model?: any, view?: any, viewModel?: any},
+    slot?: ViewSlot
+    ): CompositionContext {
+    return {
+      container,
+      bindingContext: settings.viewModel ? null : bindingContext,
+      viewResources: this.viewResources as any, // there is a TS error if this is not any ?
```
bigopon, line 236: Hmm, maybe let's solve this later. It's ok for now I guess? Did you experience any error for not having this?

packages/modal/src/resources/modal-util.ts
ben-girardet: This is a mistake (left from a previous version of this code). Thanks for catching it, will simply remove.

packages/modal/src/ux-modal-service.ts
ben-girardet: No problem found with this `any`.

packages/modal/src/ux-modal-service.ts
ben-girardet: Good idea https://github.com/aurelia/ux/pull/255/commits/08a48f25047313c259dc23f54428d4fbcd21d008

packages/modal/src/ux-modal.ts
```diff
@@ -0,0 +1,361 @@
...
+
+const log = getLogger('ux-modal');
+
+@inject(Element, StyleEngine, UxModalService, TaskQueue, UxDefaultModalConfiguration)
+@customElement('ux-modal')
+export class UxModal implements UxComponent {
+
+  @bindable public position: UxModalPosition = 'center';
+  @bindable public host: 'body' | HTMLElement | false | string = 'body';
+  @bindable public modalBreakpoint: number = 768;
```
bigopon, line 19: Probably minor. But we will need some accomodating comments for what things are later. It may not be as obvious as it seems, to anyone beside the author.

bigopon:
@ben-girardet looks good to me. I have no further comments for now. Let's go first and continue enhancing this.

PR #259, opened 2020-05-08
fix(checkbox): center ripple container

MaximBalaganskiy:
offset = 24px + (50px - 24px)/2 = 37px


bigopon:
@ben-girardet any comments on this?

MaximBalaganskiy:
image
2 pixels off

bigopon:
Ok this looks good to me, if we have issues with the new calculation, we can fix it anytime. Thanks @MaximBalaganskiy

Issue #260, opened 2020-05-08
Already extracted normalize.css is being added to global styles


MaximBalaganskiy:
**I'm submitting a bug report**

* **Library Version:**
0.20.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
12.16.2

* **NPM Version:**
6.14.4

* **Webpack Version**
webpack 4.42.1

* **Browser:**
all

* **Language:**
TypeScript 3.8

**Current behavior:**
`ux-configuration.ts` module loads `@aurelia-ux/core/styles/normalize.css` into global styles tag. 
https://github.com/aurelia/ux/blob/ed989f92fb10948f086e92a3665814444f9e2a45/packages/core/src/ux-configuration.ts#L29

But because webpack extracted it already an empty object gets stringified which results in invalid style tag. This invalidates consequently added styles.
The bug can be observed at https://ux.mintello.com/.

**Expected/desired behavior:**
The style is not added when it has been previously extracted by the loader.

* **What is the motivation / use case for changing the behavior?**
Allow components adding their own global styles.

bigopon:
We still need to include this css module, so we probably just need to tweak
```ts
PLATFORM.moduleName('./styles/normalize.css');
```
to
```ts
PLATFORM.moduleName('text-loader!./styles/normalize.css');
```
maybe?

MaximBalaganskiy:
Duplicate of #171

PR #261, opened 2020-05-08
feat(progress): add progress

MaximBalaganskiy:
Basically a copy of angular's progress spinner


packages/progress/rollup.config.js
```diff
@@ -0,0 +1,55 @@
+import typescript from 'rollup-plugin-typescript2';
+import packageJson from './package.json';
+import { copy, html, typings, buildCopyInstruction } from '../../rollup.plugins'
+
+const FILE_NAME_ENTRY = 'src/index.ts';
+const FILE_NAME_OUTPUT = 'index.js';
+const FILE_NAME_CSS = [
+  'ux-button.css'
```
bigopon, line 8:
```suggestion
  'ux-progress.css'
```

packages/progress/src/ux-progress.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.diameterNumber = normalizeNumberAttribute(this.diameter);
+    this.strokeWidthNumber = normalizeNumberAttribute(this.strokeWidth);
+    this.update();
+  }
+
+  private update() {
+    this.strokeCircumference = ((this.diameterNumber ?? 0) - (this.strokeWidthNumber ?? 0)) * 3.14;
+    if (this.valueNumber === undefined || this.valueNumber === null) {
+      this.strokeDashOffset = 0;
+      const styleId = `ux-progress-animation-template-${this.diameter}-${this.strokeWidthNumber}`;
```
bigopon, line 81: does this mean the old styleId will be left orphaned in the document?

packages/progress/rollup.config.js
MaximBalaganskiy: addressed

packages/progress/src/ux-progress.ts
MaximBalaganskiy: Unless we track which progress instance uses which style. Is it worth it?
Is there an API which does it?

bigopon:
Please revert the dist files 😬 they ll be taken care by the release process

packages/progress/src/ux-progress.ts
bigopon: Yes, we have inject style api for this. You can pass the id of the style element in 2nd argument so that it will reuse an existing one if already injected

MaximBalaganskiy:
Should they be in the repo in the first place? :P

packages/progress/src/ux-progress.ts
MaximBalaganskiy: That API is broken - https://github.com/aurelia/ux/issues/260
Btw, that API does not hold element references either.

MaximBalaganskiy:
The way the repo is configured - with the single tsconfig for vscode and build process - it will be inconvenient to not submit dist files, imho

bigopon:
Can you help elaborate why a separate tsconfig pointing to source file helps vscode?

MaximBalaganskiy:
Imagine you just cloned the repo and don't have dist files yet, or just updating the core package and dist files are outdated. Aliasing modules to src folders will fix intellisense issues for vscode.

bigopon:
@MaximBalaganskiy that sounds like a pretty good improvement. Could you help include that in this PR? I'd love it

packages/progress/src/ux-progress.ts
bigopon: That API doesn't need to hold element reference, it just need to use the same ID for the `<style/>` element every time it's called. Btw, can you please help try my suggestion in that issue? Adding a `text-loader!` in front of the module name, and load the module with only the path?

MaximBalaganskiy:
Does this look ok?

packages/progress/package.json
```diff
@@ -0,0 +1,146 @@
...
+  "aurelia": {
+    "build": {
+      "resources": [
+        "@aurelia-ux/progress/ux-progress"
+      ]
+    },
+    "documentation": {
+      "articles": []
+    }
+  }
```
bigopon, line 145: we no longer need this

packages/progress/package.json
```diff
@@ -0,0 +1,146 @@
...
+  ],
+  "homepage": "http://aurelia.io",
+  "bugs": {
+    "url": "https://github.com/aurelia/ux/issues"
+  },
+  "license": "MIT",
+  "author": "Rob Eisenberg <rob@bluespire.com> (http://robeisenberg.com/)",
+  "contributors": [
+    "Zach Hollingshead <zach.hollingshead@gmail.com>",
+    "bigopon <bigopon.777@gmail.com>"
```
bigopon, line 20: would be great if you can add your name here ❤️

packages/core/src/ux-configuration.ts
```diff
@@ -16,7 +16,7 @@ export class UXConfiguration {
   }
 
   public cssNormalize() {
-    PLATFORM.moduleName('./styles/normalize.css');
+    PLATFORM.moduleName('!css-loader!./styles/normalize.css');
```
bigopon, line 19: @3cp will this affect cli bundler in anyway? We need to prepend `!css-loader!` to make sure webpack does not extract the css (the starting `!`). Though I'm thinking maybe we just need the check below...
```ts
if (text && typeof (text) === 'string') {
```

3cp: css-loader is definitely should not be there, because it's webpack only.

I understand there is some complication to load css dependency inside js/ts file, because our standard webpack config only does skip style-loader for css dep inside html file.

```
{
        test: /\.css$/i,
        issuer: [{ not: [{ test: /\.html$/i }] }],
        use: extractCss ? [{
          loader: MiniCssExtractPlugin.loader
        }, ...cssRules
        ] : ['style-loader', ...cssRules]
      },
      {
        test: /\.css$/i,
        issuer: [{ test: /\.html$/i }],
        // CSS required in templates cannot be extracted safely
        // because Aurelia would try to require it again in runtime
        use: cssRules
      },
```

There are two options you can get away from this:
1. don't use css file, use an inline string, then use `DOM.injectStyles(cssString)` to load the style.
2. use `<require from="./style.css"></require>` in a html file.

This is necessary to support unknown bundlers, although there are currently only webpack, cli-bundler, and dumber can work with Aurelia 1.

packages/progress/package.json
3cp: Correct, it was for aurelia-cli 0.32 and below.

packages/core/src/ux-configuration.ts
MaximBalaganskiy: @3cp Would `import 'normalize.vss';` work?

3cp: Yes you are right, it should, aurelia-dialog does that.
We just need a dist build that cli-bundler can understand, cli-bundler doesn't understand `import()` (due to esprima). (BTW, dumber has no problem to support `import()`)

aurelia-dialog does that with an umd build, pointed by `"browser"` field in package.json.

3cp: But honestly, I am not sure how webpack deals with dep in `import()`. Not sure how it skipped `style-loader`.

3cp: Sorry, I totally misread your question 😆 

Yes, `import 'normalize.css'` should work, just leave it to bundler to load the style.

3cp: I don't know how ux supports style customisation. But `import 'normalize.css'` statically load the style, the `DOM.injectStyle()` used in aurelia-dialog allow a delayed injection of css, possibly customised by user.

MaximBalaganskiy:
Done

packages/core/src/ux-configuration.ts
```diff
@@ -1,8 +1,8 @@
 import { inject } from 'aurelia-dependency-injection';
 import { getLogger } from 'aurelia-logging';
 import { Loader } from 'aurelia-loader';
-import { PLATFORM } from 'aurelia-pal';
 import { GlobalStyleEngine } from './styles/global-style-engine';
+import './styles/normalize.css';
```
bigopon, line 5: it's a difficult situation. I think it's unreasonable to distribute this. Maybe lets remove this line 
```
    PLATFORM.moduleName('!css-loader!./styles/normalize.css');
```
and just have
```ts
    PLATFORM.moduleName('./styles/normalize.css');
```
That together with the check
```ts
        if (text && typeof (text) === 'string') {
```
should be enough.

MaximBalaganskiy: but this is effectively what that code does, just by calling loader api... what if a customer wants to extract css into a separate css bundle?

bigopon: Isn't it the same for the combo `PLATFORM.moduleName('...')` + `if(...)` vs `import '...'`? If it's extracted then the if block does nothing, and it doesn't require us any instrumentation for the build process. I think `import '..'` will fail current build process as well.

MaximBalaganskiy: why would it? `npm run build` runs just fine, `normalize.css` gets copied to the dist folder

MaximBalaganskiy: @bigopon @3cp @ben-girardet so, what do we do?

bigopon: Ill test it tonight or tmr and get back to this

bigopon: @MaximBalaganskiy it works ... so it will generate this 
image

I haven't used it this way for so long that I didn't even know it. So that's great, we just need to ask user to make sure they handle `import './style/normalize.css'` in their build.


Works fine with the CLI bundler too
image

So i guess this is good. Maybe we can do it for everywhere else and start bundling with Rollup for simpler distribution again.

Thanks @MaximBalaganskiy 

Do you have any comments on this @EisenbergEffect?

bigopon: btw, we may need to tweak the id of the style element injected, so that it's always valid (i'm not entirely sure if the `_au_css:@....` is a valid id

bigopon:
For anyone who wants to have a play/look at the component https://vercel.com/bigopon/aurelia-ux-demo/hb9j8gszp (scroll to bottom)

This looks good to me, so only waiting for @EisenbergEffect 's comment for the `.css`

MaximBalaganskiy:
Imports like this have already been used - see input component.

EisenbergEffect:
@bigopon Let's go with this, and then maybe follow-up to see if we can use the same approach throughout the codebase.

bigopon:
Ok, sounds good then. Let's go

PR #263, opened 2020-05-11
feat(sidenav): add sidenav

MaximBalaganskiy:


bigopon:
Some animated gif would be nice, please 😁 

Optionally, you can build the demo app, and publish it some where like what @ben-girardet did

MaximBalaganskiy:
https://maximbalaganskiy.github.io/aurelia-ux-demo

ben-girardet:
Thanks @MaximBalaganskiy 
Cool to see the lib moving forward with new components :-)

What's your plan regarding mobile support for the sidenav ? Material Design recommend a 600px breakpoint for the sidenav to go modal. Suggestion here would be to have a bindable breakpoint. When the breakpoint is reached the sidenav goes fullscreen. What do you think ?

Also, what about adding a `side="bottom"` (also to follow Material guidelines) ?

MaximBalaganskiy:
Honestly, I haven't even looked at the guidelines :) Will do tomorrow

MaximBalaganskiy:
@ben-girardet At the moment, the modal is achieved by setting both over and backdrop attributes.
Not sure what to do here. What if a consumer wants to control when a sidenav is hidden or not?
E.g. in our projects the menu is always hidden, even on bigger screens.

Could you please also elaborate on technical details of "a bindable breakpoint"?

ben-girardet:
@MaximBalaganskiy thanks for the bottom drawer !

Regarding the modal breakpoint there might be several technical approaches. The end goal is to follow Material guidelines with a drawer that always go modal when the window reaches a certain breakpoint.

Technically I see something similar to what I have done with the `ux-modal` with a `@bindable modalBreakpoint: number = 768` property in the `ux-sidenav-drawer` component. 

Then the methods `isOver()` and `isModal()` should return `true` when the breakpoint has been reached regardless of the `over` and `modal` values. Of course there also must be a listener on the `resize` event of `window`.

Notes:
- Material recommends a `600` breakpoint. I think it is better than the `768` that I have used in the modal component.
- Folds who don't want this behavior can set a `0` breakpoint
- I also suggest to make the `ux-sidenav` plugin configurable where folks can set defaults values for such things, even for the `over` and `modal` bindables. Many application have the same behavior for multiple instances so it's very useful to be able to configure things once in the plugin and have relevant defaults throughout the whole app.

MaximBalaganskiy:
Should it observe window size or a parent size?

ben-girardet:
Good question. I think it should observe the parent size.

BTW: we have a WIP on a `size` custom attribute for listening to an element size. This is destined to be part of `core` and available for all components. The work has been done to optimize the `grid` component because for now it listen to window size and this is not good behavior and we want to change it to listen to parent size.

Here are the related files if you want to have look:

`size` attribute: 
* https://github.com/aurelia/ux/pull/241/files#diff-49ebb23b0fa7398a5327f64304e88245

`grid` files:
* https://github.com/aurelia/ux/pull/241/files#diff-e8fca0d6566ad4dea13532334293b2b9
* https://github.com/aurelia/ux/pull/241/files#diff-c7db05afbe42bd0b58014ea351a29e50

Unfortunately this work is part of a larger PR that I haven't had time to work on recently. If this can help I can take some time today to extract the `size` attribute from this PR and create a smaller one to make it available on `core`. Except if you want to take care of this ore take another path...

Maybe @bigopon has some comments too ?

MaximBalaganskiy:
@ben-girardet You did not use `ResizeObserver` because of the compatibility with IE?

ben-girardet:
We had a discussion with @bigopon about this [from this post onward](https://github.com/aurelia/ux/pull/241#issuecomment-600374671).

The idea is to use `ResizeObserver` and allow users to polypill it if needed. Because `ResizeObserver` is not available on safari I didn't feel good about just using it and ask people to provide a polyfill. This would mean that the component would not work "out of the box" on safari. Hence the `size` custom attribute is an attempt at:

* Relying on `ResizeObserver`
* Consumer can provide their own polyfill of choice if needed
* If not polyfill is provided, the lib has a fallback with dirty checking as a last resort solution to ensure it works "out of the box"

ben-girardet:
BTW: Safari just added support for `ResizeObserver` in their 13.1 release. The coverage is still quite small but we might soon be able to only rely on `ResizeObserver`

MaximBalaganskiy:
It would be great having it in core

ben-girardet:
If you feel comfortable the quickest way to get it and make it available for your PR is that you copy over the `size-attribute.ts` file in your branch and start using right away. 

https://github.com/aurelia/ux/blob/924cdf33bd3454798d390c3dcfbff6cdb76ad0aa/packages/core/src/components/size-attribute.ts

Once your PR will be merged I will rebase my branch with your additions.

MaximBalaganskiy:
Sure

MaximBalaganskiy:
@ben-girardet done. Also added some typings for `au` property

app/src/components.html
```diff
@@ -1,65 +1,109 @@
...
+        <p>
+          <ux-button class="ux-button--accent" type="outline">Outline</ux-button>
+        </p>
+        <p>
+          <ux-button class="ux-button--accent" type="fab">Fab</ux-button>
+        </p>
       </ux-grid-cell>
       <ux-grid-cell sm="4">
         <h4>Disabled (disabled)</h4>
-        <p><ux-button disabled type="raised">Raised</ux-button></p>
```
bigopon, line 30: Well, there's merit with both type of formatting, so I'd suggest we get @ben-girardet input on this. Once this view gets big, it's not too bad for merging one child block content into a single line.

packages/core/src/components/html-attributes.ts
```diff
@@ -19,3 +19,13 @@ export function normalizeBooleanAttribute(attributeName: string, value: boolean
 
   return ret;
 }
+
+export function normalizeNumberAttribute(val: number | string): number | undefined | null {
+  if (val === undefined || val === '' || val === 'undefined' || val === 'NaN') {
```
bigopon, line 24: Can you give a bit background on the comparison against `'undefined'` and `'NaN'`

app/src/components.html
MaximBalaganskiy: It's Beautify at work. The project probably needs a config checked in if formatting is important.

packages/core/src/components/html-attributes.ts
MaximBalaganskiy: There is no background. Can be removed

app/src/components.html
bigopon: It's not that important, at least for now. But since @ben-girardet creates this, I just didn't want to rock it 😓

packages/core/src/components/html-attributes.ts
MaximBalaganskiy: @bigopon what's your suggestion?

bigopon: Lets start with this. I think if it comes from your realworld usage then its fine to start with. If it proves to be against common expectation, we can change it

app/src/components.html
ben-girardet: I don’t think it’s important either. 

However as a best practice we can try to avoid formatting changes in PR as it doesn’t refer to the subject of the PR and make it a little more complicated to review. 

No need to change anything now from my point of view. Dropping this comment as a best practice idea for future commits.

ben-girardet:
@MaximBalaganskiy thanks for the changes.

Regarding the modalBreakpoint I don't see how it currently affects the behavior of the drawer. Am I missing something ?

The way I see it is a breakpoint that forces the `isOver()` and `isBackdrop()` to return true.

Currently you have

```ts
isOver(): boolean {
  return normalizeBooleanAttribute('over', this.over);
}
isBackdrop(): boolean {
  return normalizeBooleanAttribute('backdrop', this.backdrop);
}
```

could look like:

```ts
isOver(): boolean {
  return this.currentWidth < this.modalBreakpoint || normalizeBooleanAttribute('over', this.over);
}
isBackdrop(): boolean {
  return this.currentWidth < this.modalBreakpoint || normalizeBooleanAttribute('backdrop', this.backdrop);
}
```

What do you think ?

MaximBalaganskiy:
It is decided in UxSidenav class, drawers are not aware of the backdrop

ben-girardet:
Ok I missed a part in the `UxSidenav` class. 

I see how it's meant to affect the backdrop. Is it also supposed to affect the `over`?

_PS: is your demo up to date with the latest version ?_

MaximBalaganskiy:
It affects backdrop and content padding.
No, the app is outdated.

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
```diff
@@ -0,0 +1,107 @@
...
+      newValue.themeKey = 'sidenav';
+    }
+
+    this.styleEngine.applyTheme(newValue, this.element);
+  }
+
+  attached() {
+    this.leftDrawer = this.element.querySelector<IUxSidenavDrawerElement>('ux-sidenav-drawer[side="left"]')?.au.controller.viewModel;
+    this.rightDrawer = this.element.querySelector<IUxSidenavDrawerElement>('ux-sidenav-drawer[side="right"]')?.au.controller.viewModel;
+    this.bottomDrawer = this.element.querySelector<IUxSidenavDrawerElement>('ux-sidenav-drawer[side="bottom"]')?.au.controller.viewModel;
```
ben-girardet, line 42: Could we make this more flexible ? I guess that discovering the drawers on `attached` will be fine for 95% of usages but do not cover edge cases where one would want to have drawer elements behind `if.bind` ?

Wouldn't it be better to use `@processContent` of `@children` decorators to discover the drawers ?

@bigopon any thoughts ?

packages/sidenav/src/ux-sidenav-drawer/ux-sidenav-drawer.scss
```diff
@@ -0,0 +1,45 @@
+@import '../common.scss';
+
+ux-sidenav-drawer {
+  position: absolute;
+  z-index: 3;
+  background-color: var(--aurelia-ux--design-control-background, white);
+  color: var(--aurelia-ux--design-control-foreground, black);
```
ben-girardet, line 7: I think it would be better to have a themable properties for background/foregound of the drawer and if nothing provided defaults to `surface` pair.

```css
background-color: var(--aurelia-ux--drawer-background, var(--aurelia-ux--design-surface-background, #F5F5F5));
color: var(--aurelia-ux--drawer-foreground, var(--aurelia-ux--design-surface-foreground, #212121));
```

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
MaximBalaganskiy: It works under if.bind, check out theming page. What does NOT work is @child under if.bind. It never gets populated

MaximBalaganskiy:
As for scrolling, I think that should be left to a consumer. I personally prefer perfect-scrollbar which is very unobtrusive and easy to use

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
bigopon: I'll be fixing this one over this weekend. For now, this is probably the way we have to do it 😓

packages/sidenav/src/ux-sidenav-drawer/ux-sidenav-drawer.scss
MaximBalaganskiy: Done

ben-girardet: Cool thanks. As defaults I would use the `surface` pair instead of the `control` pair. But you might have a reason why you defaults to control ?

MaximBalaganskiy: Coz I've found them first :)

MaximBalaganskiy: Actually, no... With surface colours the drawers will blend with surroundings

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
ben-girardet: You mean that it will work with this ? 

```html
<ux-sidenav class="theming">
  <ux-sidenav-drawer side="left" if.bind="shouldDisplayDrawer">Drawer</ux-sidenav-drawer>
  <!-- ... -->
</ux-sidenav>
```

Didn't see this in the theming page.

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
```diff
@@ -0,0 +1,107 @@
...
+export class UxSidenav {
+  constructor(public element: HTMLElement, private styleEngine: StyleEngine, private taskQueue: TaskQueue,
+    private defaultConfig: UxDefaultSidenavConfiguration) { }
+
+  backdrop: boolean = this.defaultConfig.backdrop;
+  width: number;
+
+  leftDrawer: UxSidenavDrawer | undefined;
+  rightDrawer: UxSidenavDrawer | undefined;
+  bottomDrawer: UxSidenavDrawer | undefined;
```
ben-girardet, line 20: Another edge case that I can imagine... 

A situation where the consumer needs several `left` (or any other position) drawers. From what I understand in `ux-sidenav.ts` it would be possible to achieve this with having an array of `drawers` instead of 3 distincts properties for `leftDrawer`,  `rightDrawer` and `bottomDrawer`?

MaximBalaganskiy: ah, this... no, not supported :)

MaximBalaganskiy: can enhance later... i'd like to merge all three branches first, it's getting harder to juggle them

ben-girardet:
Thanks again for your work ! Left another comment... just trying to find edge cases. We can still discuss what we want to cover immediately and what can be done later...

Can't wait to use the comp

packages/sidenav/src/ux-sidenav-drawer/ux-sidenav-drawer.scss
ben-girardet: In the theming page the components are demoed inside a card -> which is a surface. So yes, in this scenario it will blend with surroundings and that's where you need a `theme.bind="{}"` on your sidenav to change this style.

To be honest I don't think that this demo inside a card is a great design. But it's what we have yet (and it's my fault btw). Hopefully we'll have a better demo page soon ;-)

But by design a sidenav is more surface than a control. A sidenav should not be used inside card anyway ?!?

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
ben-girardet: I'm ok with this.

MaximBalaganskiy: I don't insist, happy to change it

packages/sidenav/src/ux-sidenav-drawer/ux-sidenav-drawer.scss
MaximBalaganskiy: Changed

packages/core/src/components/i-controller.ts
```diff
@@ -0,0 +1,3 @@
+export interface IController<T> {
```
bigopon, line 1: I'm for having interfaces in a separated file for discover-ability, but a file per interface could be too wild. Let's group all interfaces in a file name either `definitions` (v2) or `interfaces` (v1)

bigopon:
A playable demo for anyone who wants to have a look https://aurelia-ux-demo-sidenav.now.sh/#/components

bigopon:
Ok while there's some re-organisation that could be done, we probably can do it later

packages/sidenav/src/ux-sidenav-drawer/ux-sidenav-drawer.ts
```diff
@@ -0,0 +1,41 @@
+import { useView, customElement, PLATFORM, inject, bindable } from "aurelia-framework";
+import { normalizeBooleanAttribute } from "@aurelia-ux/core";
+import { UxDefaultSidenavConfiguration } from "../ux-default-sidenav-configuration";
+
+export const OPEN_CHANGED = 'open-changed';
```
bigopon, line 5: Can this be either `OPEN_CHANGED_EVENT` on the class, or in an shared exported enum

packages/sidenav/src/ux-sidenav/ux-sidenav.ts
bigopon: here is the PR that fixes this issue https://github.com/aurelia/templating/pull/685
@ben-girardet yes, it's a subtle bug, only when you put the if on the parent and hide/show via an if.

bigopon:
Ok looks good to me. Thanks @MaximBalaganskiy

PR #264, opened 2020-05-12
fix(all): separate dev and build tsconfigs

MaximBalaganskiy:
This allows for removing dist files from the repository.
VS code picks up definitions straight from src folders.
The package tsconfigs just reference the higher level ones.

In addition, packages now explicitly import templates which removes the need to use ModuleDependenciesPlugin.


bigopon:
This seems beneficial to development. Though we may have issues with folks who rely on dist on master branch

cc @fkleuver @EisenbergEffect

3cp:
Generally, dist files should be ignored in git repo.

If user uses git repo as the dependency in their package.json, the missing dist files is not a problem, as long as this repo's package.json has proper npm script `"prepare": "npm run build"`.
(The changes in this PR is too big for me to confirm whether there is a "prepare" script.)

When npm install a git repo dep (not a normal npm dep) which has a "prepare" script, it will run it (install dev deps before running it too) to build dist files in our case.

**But yarn and pnpm did not cover this npm behaviour, mainly due to lack of understanding when they were first designed.**
https://github.com/yarnpkg/yarn/issues/5235
https://github.com/pnpm/pnpm/issues/855

BTW, yarn team ignored my fix. https://github.com/yarnpkg/yarn/pull/6131

MaximBalaganskiy:
dist files are built before publish, so npm packages will have them.
Intra repo development will use aliases to src folder.
It is arguable if one should develop an external package and reference dist file from another dev folder.

fkleuver:
> Though we may have issues with folks who rely on dist on master branch

That's the thing. I've had to completely revert a bunch of CI work because of this before and this is the reason we have this convoluted develop->master merge and merge-back process in some of the au1 repos.

Getting rid of dist files on master *is* a breaking change for some.

It's not a breaking change that I am against, though, because the benefits would be significant for our development efforts, but I don't know how many people rely on this and how hard a sell it would be for them.

MaximBalaganskiy:
It would be interesting to see a scenario where dist in master is needed.

fkleuver:
> It would be interesting to see a scenario where dist in master is needed.

Several years ago it was fairly common to install dependencies into the package.json or jspm config as a github url.
It's less common now that npm has matured more, but apparently a not insignificant number of legacy projects still rely on this.

fkleuver:
Again, I'm not against it per se, but we cannot just wipe it under the rug and pretend it's not a breaking change, because it is. And if we make this breaking change, we ought to do it for all Aurelia repos so we can simplify everything in the same way.

3cp:
> install dependencies into the package.json or jspm config as a github url

I mentioned "prepare" script in package.json can support that use case. Npm will build dist files during installation.

Removing dist files, add "prepare" script. It's only a breaking change for yarn/pnpm users. Not a breaking change for npm users.

fkleuver:
> It's only a breaking change for yarn/pnpm users.
And JSPM users, which I believe is the majority of people who use it this way.
.. a breaking change for some, is still a breaking change

3cp:
Right, jspm users (I totally ignored them) :)

MaximBalaganskiy:
So, as an alternative, accept the tsconfigs, keep the dist, and somehow setup a hook to reset dist upon commit. Should be possible...

MaximBalaganskiy:
Ok. The following hook will void any changes to dist before commit.
Either add it to your .git/hooks folder or remember to do it manually

pre-commit
```bash
#!/bin/sh

git reset -- packages/**/dist/*.*
git checkout -- packages/**/dist/*.*
```

MaximBalaganskiy:
Does this look ok now?

packages/button/src/ux-button.ts
```diff
@@ -1,124 +1,126 @@
...
-}
+import { customElement, bindable, useView } from 'aurelia-templating';
+import { inject } from 'aurelia-dependency-injection';
+import { StyleEngine, UxComponent, PaperRipple, normalizeBooleanAttribute } from '@aurelia-ux/core';
+import { UxButtonTheme } from './ux-button-theme';
+import { PLATFORM } from 'aurelia-framework';
+
+@inject(Element, StyleEngine)
+@customElement('ux-button')
+@useView(PLATFORM.moduleName('./ux-button.html'))
```
bigopon, line 9: why is this needed?

bigopon: I remember we did this, but then had to move back to the original approach with `PLATFORM.moduleName` in index file to ensure compatibility with normal app, and our demo app

MaximBalaganskiy: Don't remember where exactly I saw it, but a team member posted that all plugin developers should explicitly reference templates so that webpack bundles them without the need for `ModuleDependenciesPlugin`

MaximBalaganskiy: bundling an external app, not the demo one, does not pick up templates without this

bigopon: When building the dependency graph of a module, in `index` file, if we do:
```ts
  .globalResources(PLATFORM.moduleName('./ux-button'))
```

Aurelia webpack plugin is taught to start looking for the corresponding html module of `ux-button` module as well, so we don't need to do `PLATFORM.moduleName('ux-button.html')` inside `ux-button` view model module.

If you do:
```ts
import { UxButton } from './ux-button';

// index.ts of ux-button module
  .globalResources(UxButton);
``` 
Then it's a different story. Because there is no `PLATFORM.moduleName` here to tell webpack that when it add `ux-button` as a dependency of `index` module, it should also find corresponding `ux-button.html` and add it as well. Hence the advice.

The above is what I'm guessing based on your comment. It may not be exactly what the other team member said, but it wouldn't be too far off. For that, plus the fact that we do `PLATFORM.moduleName` in our `index` modules, we don't need to double declare the html module of the view model module.

bigopon: Though, there's a case where what you described is beneficial: someone may want to use `UxButton` directly instead of using `.plugin` api:

```ts
import { UxButton } from '@aurelia-ux/button';

export function configure(aurelia) {
  aurelia.use.globalResources(UxButton);
}
```
I think we can try to support than, but let's discourage this usage for now, or at least have better testing story for this first

MaximBalaganskiy: It comes back now... yes it was said in that context. But there is another scenario - multi project repositories with root `node_modules`. By default, the plugin searches for template files in the local `node_modules`. We need to either pass an additional parameter to it, or a plugin needs to explicitly reference the templates :)

bigopon: Ok, that sounds good to me. And given that this change works in all scenarios: demo app, normal app and app nested within a root `node_modules`, I think it's good to go.

bigopon:
@MaximBalaganskiy thanks for the PR 👍

ben-girardet:
Looking forward to try this 👍🏻

PR #269, opened 2020-05-22
clean(dependencies)

MaximBalaganskiy:
Removed aurelia references from package dependencies except the core package.
Moved core reference to peer dependencies.
Core package has commonly used aurelia references in peer dependencies.
Packages like moment or event-aggregator are added directly via peer dependencies.
All peer dependencies are installed in the root folder.
Unified build scripts in modal, positioning.


packages/modal/package.json
```diff
@@ -19,7 +19,7 @@
   ],
   "main": "dist/commonjs/index.js",
   "module": "dist/native-modules/index.js",
-  "typings": "dist/commonjs/index.d.ts",
+  "typings": "dist/types/index.d.ts",
```
bigopon, line 22: 👍

bigopon:
@EisenbergEffect @fkleuver so I think this could be a good change. Though after this, installation requires explicit for both core and the component package. and core becoming the peer dependency of components could give less headache in users app too. Do you have any thoughts that this may not be a good idea?

cc @3cp

bigopon:
Ok let's go with this and see. If it proves to be a pain for normal usage then we will move it back.

PR #272, opened 2020-05-23
clean(lookup): dependencies

MaximBalaganskiy:
Add core and input to dev deps to hoist them via lerna. Otherwise build won't find the dist folders.
Remove aurelia dependencies - they are "peered" through core.
Unify other dev dependencies.


bigopon:
Thanks

PR #273, opened 2020-05-24
chore: setup tests & circle CI

bigopon:


bigopon:
@MaximBalaganskiy @ben-girardet after this PR, tests will start working in normal PR flow.

PR #276, opened 2020-05-25
fix(icons): use theme foreground

MaximBalaganskiy:


bigopon:
it feels `currentColor` is intentional, for example: maybe different icon colors for different sections of an app? What is the issue that you ran into?

MaximBalaganskiy:
The theme property does nothing at the moment. To change the icon colour I have to set it directly via CSS. While I have no problems doing that, it seems that not passing a theme colour is an oversight.

ben-girardet:
`currentColor` is very important as it makes the SVG automatically adjust to the current color in which the icon is used. For example, when using the `ux-icon` inside a `ux-button` the icon directly takes the current foreground color from the button.

I've found that with this `currentColor` property I rarely need to adjust the color of the icon myself.

It's true however that it doesn't make sense to have a `foreground` CSS property and not using it in the icon `.css`. 

Seems like the most reasonable thing is to defaults the `var(...)` to `currentColor`. I have the feeling that this was not done previously as CSS variables were not well enough supported, maybe esp. in the SVG context. I would be quite confident nowadays. Hopefully doesn't break something ?

MaximBalaganskiy:
fixed

PR #277, opened 2020-05-25
fix(core): rename size attribute to ux-size

MaximBalaganskiy:
Otherwise it masks `ux-icon.size` attribute.
In addition, `size` is quite popular and might conflict with consumer attributes


bigopon:
@ben-girardet you may need to be aware of this, since I think you've been using `size` in your demo app/apps. I think this is quite reasonable so i'll go with this, similar to how we went with `ux-attach-focus`.

ben-girardet:
> @ben-girardet you may need to be aware of this, since I think you've been using `size` in your demo app/apps. I think this is quite reasonable so i'll go with this, similar to how we went with `ux-attach-focus`.

Yes, no problem, I’m using it in the WIP demo PR. It’s a good idea to rename this attr.  Thanks for the heads up!

PR #279, opened 2020-05-25
Feat/add focus api

bigopon:
close #227

### DIFF
--- a/design-decisions.md
+++ b/design-decisions.md
@@ -1,0 +1,35 @@
+# `focus` attribute
+
+*You can contribute to the dicussion on this thread: https://github.com/aurelia/ux/issues/170*
+
+### API
+
+Inputable components should implement a `focus()` and `blur()` public API. Calling these methods must
+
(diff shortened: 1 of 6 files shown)


bigopon:
@ben-girardet i've cherry-picked a few commits in your PR here #227 (input/textarea)

For more complex components like chipinput/datepicker in that PR, I think a fresher implementation would be better. Also i've changed from altering the prototype of the host element to only defining what's needed to work like a native element (it's kind of a bad idea offending optimization via altering the prototype). The new modification should work fine.

ben-girardet:
thanks @bigopon 

I'm happy to go this way and input/textarea are definitely the most important components requiring this focus api. I remember that it made a big difference when using these components in a Cordova application. 

If you are confident with this PR I suggest you can merge this cherry-picking in master. I'll test the next release in some cordova projects and see how it goes.

ben-girardet:
Thanks for your help with the PR. It's good to see the number of opened PR decreasing :-)

PR #281, opened 2020-05-28
fix(input-info): add missing error classes

MaximBalaganskiy:


bigopon:
Thanks @MaximBalaganskiy

PR #282, opened 2020-05-29
feat(tree-view): add tree view

MaximBalaganskiy:


packages/tree-view/src/ux-tree-view/ux-tree-view.html
```diff
@@ -0,0 +1,22 @@
+<template class="ux-tree-view">
+  <require from="./ux-tree-view.css"></require>
+  <ul class="ux-tree-view--list">
+    <li repeat.for="$node of nodes" class="ux-tree-view--list-item">
+      <let $tree.bind="element"></let>
+      <let $tree-vm.bind="$parent"></let>
+      <div class="ux-tree-view--node ${$node.selected ? 'ux-tree-view--node-selected' : ''}" click.delegate="nodeClicked($node)">
+        <ux-button if.bind="$node.children" type="fab" size="small" class="ux-tree-view--expander"
+          click.delegate="toggleExpanded($node, $event)">
```
bigopon, line 9: let's use `.trigger`, `.delegate` should only be used in app, as it could be interfered by shadowdom. Though we fixed it, it still feels better encapsulated with `.trigger`

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
```diff
@@ -0,0 +1,118 @@
...
+const getNextNodeTemplateId = () => ++id;
+
+@inject(Element, Container, StyleEngine, UxDefaultTreeViewConfiguration)
+@customElement('ux-tree-view')
+@useView(PLATFORM.moduleName('./ux-tree-view.html'))
+@processContent(UxTreeView.processContent)
+export class UxTreeView implements UxComponent {
+  static processContent(_viewCompiler: ViewCompiler, _resources: ViewResources, element: Element, _instruction: BehaviorInstruction) {
+    const treeNode = element.querySelector('ux-tree-node');
+    if (treeNode) {
```
bigopon, line 21: I'm wondering if we should have this logic:
- if there is a `<tree-node/>`, then we use the content of that as template
- else, we use all the children of `<tree-view/>` as the content of the inner `tree-node` template
Probably it can help reduce the boilerplate of having to write `<tree-node/>`, though it would be really ambiguous as you don't know what the content is for

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
```diff
@@ -0,0 +1,118 @@
...
+    return false;
+  }
+
+  nodeClicked(n: INode) {
+    if (this.selectedNode) {
+      this.selectedNode.selected = false;
+    }
+    n.selected = true;
+    this.selectedNode = n;
+    this.element.dispatchEvent(new CustomEvent(UxTreeView.NODE_SELECTED_EVENT, { detail: { node: n }, bubbles: true }));
```
bigopon, line 85: Can we have this into multi-lines?

bigopon: A potential benefit of chopping it into multiple lines is: we may want to pass more information to the detail object in the future, where the current tree-view of the node may be needed

packages/tree-view/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import { FrameworkConfiguration, PLATFORM } from 'aurelia-framework';
+import { UxDefaultTreeViewConfiguration } from './ux-default-tree-view-configuration';
+import { UxTreeView } from './ux-tree-view/ux-tree-view';
+
+export function configure(config: FrameworkConfiguration, callback?: (config: UxDefaultTreeViewConfiguration) => void) {
+  config.globalResources([
+    PLATFORM.moduleName('./ux-tree-view/ux-tree-view'),
+    PLATFORM.moduleName('./ux-tree-node/ux-tree-node')
```
bigopon, line 8: is `ux-tree-node` supposed to be a global custom element? if not we can move this line to `<require/>` in tree view, as it seems currently we are only using it as a pseudo custom element

bigopon:
This is awesome 😁 
Can you help give some visualization please. Maybe a link?

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
MaximBalaganskiy: I considered this. I think it's better to be explicit here

packages/tree-view/src/ux-tree-view/ux-tree-view.html
MaximBalaganskiy: done

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
MaximBalaganskiy: done

packages/tree-view/src/index.ts
MaximBalaganskiy: done

MaximBalaganskiy:
https://maximbalaganskiy.github.io/aurelia-ux-demo/#/components

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
bigopon: Yeah, maybe it's best to do that here.

packages/tree-view/src/ux-tree-view/ux-tree-view.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  expandPath(path: number[]) {
+    if (path.length === 1) {
+      this.nodeClicked(this.nodes[path[0]]);
+      this.element.querySelectorAll('.ux-tree-view--node')[path[0]].scrollIntoView();
+    } else {
+      this.nodes[path[0]].expanded = true;
+      // let Aurelia populate treeViews by queueing the task
+      this.taskQueue.queueTask(() => {
```
bigopon, line 123: if there's a search that returns 10 nodes in 10 different subtrees, doesn't this mean the last one will be focused?

MaximBalaganskiy: When a node found in the tree the path is returned immediately, so the first occurrence will be focused.

bigopon: I see. Should:
- it allow multiple paths return in `findPath`, say you have 10 departments, each department is a tree. and you want to search for a person name John in all those departments, limiting to a first department only is quite limited.
- this block
  ```ts
  if (path.length === 1) {
      this.nodeClicked(this.nodes[path[0]]);
  ```
  seems to be always click on the first path gradually, can you help explain why it's needed? Shouldn't that cause unwatned side effect - example: parent - child view where intermediate items would come into focus unnecessarily?

MaximBalaganskiy: The path array is a list of node indexes which need to be expanded. The last index is the one we're looking for

MaximBalaganskiy: as for multiple matches, didn't really think about that as the only requirement for me was to search by id which is unique in my case. Even if you have a name search, it is natural to focus the first occurrence

bigopon: yes, the last index will be focused, eventually. Though I think by faking click on the intermediate nodes, we will run into side effect issue, hypothetically like I mentioned before. Can we do the expand from code?

For multiple matches, I meant we are supposed to show in all subtrees, not just only the first one. Example:
```
company
  | department 1
      | - John Wick
  | department 2
      | - John Wicked
  | department 3
      | - John Wicker
  | department 4
      | - John Wickest
```
Searching for John should open the tree for 4 departments, but only focus on the first one.

MaximBalaganskiy: but the intermediate nodes are not clicked, only the last one

MaximBalaganskiy: as for multiple search - quite easy to do in flat mode, but a bit harder in hierarchy. I'd rather removed the API to let consumers do it.

bigopon: > but the intermediate nodes are not clicked, only the last one

Wait i see it. Yes it's is only clicking on the last one. I read the code wrong, sorry.

bigopon: > as for multiple search - quite easy to do in flat mode, but a bit harder in hierarchy. I'd rather remove the API to let consumers do it.

Ok lets get this in for now, I'll do an enhancement later when we need it

bigopon:
@MaximBalaganskiy is this ready to go, what else left todo?

MaximBalaganskiy:
I had no more plans for this atm

packages/tree-view/src/ux-default-tree-view-configuration.ts
```diff
@@ -0,0 +1,5 @@
+import { UxTreeViewTheme } from './ux-tree-view/ux-tree-view-theme';
+
+export class UxDefaultTreeViewConfiguration {
+  public theme?: UxTreeViewTheme;
```
ben-girardet, line 4: I kinda like this idea to have the theme in the configuration. But giving it more thoughts I'm afraid it leads consumers the wrong way... Just want to highlight something here to discuss the "default style scenario".

Without the `theme` configuration, the original way to apply a default style to a component is through the core `styleEngine.applyTheme()` method. When we use this method without passing an `element` then the styles are applied as default in the `<HEAD/>`.

Now I see two benefits from using `applyTheme()`:

1. I think it has better perf as it only apply the theme once
2. More subtle: in this case when we pass a `theme` to a component, it will only override the default theme. On the contrary, with the default `theme` property on the plugin, the component will either use the default, or use the binded property.

My conclusions:
- I like to have the `theme` property as a configuration as it makes it super easy for consumers to set a default theme
- I'm not so confident with the idea to have a either/or scenario for the default theme. I think that if I set a nice button theme as a default but for some reason I need to change only the color at one place, I would prefer to only pass `theme.bind="{background: 'red'}"` instead of a whole new theme
- Seems to me that perf-wise the `applyTheme()` is better

What do you think ? Should we keep the `theme` configurable option but in the `configure()` we do a `applyTheme()` (as a postTask or something like that) instead of applying the default theme in the component constructor ?

packages/tree-view/src/ux-tree-view/i-ux-tree-view-element.ts
```diff
@@ -0,0 +1,9 @@
+import { IController } from '@aurelia-ux/core';
+import { UxTreeView } from './ux-tree-view';
+
+export interface IUxTreeViewElement extends HTMLElement {
+  au: {
+    controller: IController<UxTreeView>;
+    'ux-sidenav': IController<UxTreeView>;
```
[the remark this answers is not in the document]
MaximBalaganskiy: nope :)

packages/tree-view/src/ux-default-tree-view-configuration.ts
MaximBalaganskiy: It does create ambiguity. I would remove it. But leave properties like dense or variant and pass them to inputs

packages/tree-view/src/ux-tree-view/i-ux-tree-view-element.ts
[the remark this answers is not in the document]
MaximBalaganskiy: fixed

packages/tree-view/src/ux-default-tree-view-configuration.ts
MaximBalaganskiy: I would even remove it from the config as long as StyleEngine is documented

packages/tree-view/src/ux-tree-view/i-ux-tree-view-element.ts
```diff
@@ -0,0 +1,9 @@
+import { IController } from '@aurelia-ux/core';
+import { UxTreeView } from './ux-tree-view';
+
+export interface IUxTreeViewElement extends HTMLElement {
+  au: {
```
bigopon, line 5: Do we need this?

bigopon:
Thanks @MaximBalaganskiy for the initial implementation 👍

packages/tree-view/src/ux-tree-view/i-ux-tree-view-element.ts
MaximBalaganskiy: Not at the moment, just for future needs

PR #285, opened 2020-06-02
fix(lookup): add class for validation renderer

MaximBalaganskiy:


bigopon:
is there any reason the validation couldn't use `ux-lookup`? If we do this, we need to do it for every component.

MaximBalaganskiy:
The renderer in the demo app relies on these classes. E.g. ux-input already has it

ben-girardet:
I think we already do this for the other components. Plus I think it's very good to use BEM styles which requires us to declare the `.ux-lookup` class in the element.

bigopon:
Cool, thanks for confirming @MaximBalaganskiy @ben-girardet lets go

PR #286, opened 2020-06-02
fix(input): pass through date input types, default config

MaximBalaganskiy:
Returning true is needed to support clicks on a native calendar button.

### DIFF
--- a/packages/input/src/ux-input.ts
+++ b/packages/input/src/ux-input.ts
@@ -173,6 +173,9 @@ export class UxInput implements UxInputComponent {
   public typeChanged(newValue: any) {
     if (![
       'text',
+      'date',
+      'time',
+      'datetime-local',
       'password',
       'number',
       'email',
@@ -197,6 +200,7 @@ export class UxInput implements UxInputComponent {
 
   public focus() {
     this.textbox.focus();
+    return true;
   }
 
   public blur() {
--- a/packages/input/src/ux-input.ts
+++ b/packages/input/src/ux-input.ts
@@ -46,16 +46,16 @@ export class UxInput implements UxInputComponent {
   public value: any;
   public textbox: HTMLInputElement;
 
-  constructor(private element: UxInputElement, public styleEngine: StyleEngine, private defaultConfiguration: UxDefaultInputConfiguration) {
+  constructor(private element: UxInputElement, public styleEngine: StyleEngine, defaultConfiguration: UxDefaultInputConfiguration) {
     defineUxInputElementApis(element);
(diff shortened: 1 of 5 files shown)


bigopon:
I guess there's no reason to not pass this through. Though do we need to handle the value of date-* input differently? always get as Date object or just a string?

MaximBalaganskiy:
Converting to a date would be convenient but not consistent with the native input. Not sure, personally

bigopon:
> Converting to a date would be convenient but not consistent with the native input. Not sure, personally

we have `input.valueAsDate` for this, though I've never tried all `input`/`change` events combination with different browsers to be sure enough of this either.

packages/input/src/ux-input.ts
```diff
@@ -8,12 +8,15 @@ import { UxInputTheme } from './ux-input-theme';
...
 // tslint:disable-next-line: no-submodule-imports
 import '@aurelia-ux/core/components/ux-input-component--outline.css';
+import { UxDefaultInputConfiguration } from './ux-default-input-configuration';
 
 export interface UxInputElement extends HTMLElement {
   value: any;
 }
 
-@inject(Element, StyleEngine)
+export type InputVariant = 'filled' | 'outline';
```
bigopon, line 17: maybe this type should be exported from core, as we have `InputComponent` from there

packages/input/src/ux-input.ts
```diff
@@ -43,8 +46,17 @@ export class UxInput implements UxInputComponent {
   public value: any;
   public textbox: HTMLInputElement;
 
-  constructor(private element: UxInputElement, public styleEngine: StyleEngine) {
+  constructor(private element: UxInputElement, public styleEngine: StyleEngine, private defaultConfiguration: UxDefaultInputConfiguration) {
```
bigopon, line 49: if the configuration is not used anywhere else after the constructor, let's not assign it to the ux input

packages/input/src/ux-input.ts
```diff
@@ -43,8 +46,17 @@ export class UxInput implements UxInputComponent {
...
     defineUxInputElementApis(element);
+    if (this.defaultConfiguration.theme !== undefined) {
```
bigopon, line 51: @ben-girardet we now have default configuration for the ux input as well

packages/input/src/ux-input.ts
MaximBalaganskiy: done

packages/input/src/ux-input.ts
MaximBalaganskiy: done

PR #287, opened 2020-06-02
fix(select): add default configuration

MaximBalaganskiy:


packages/select/src/ux-select.ts
```diff
@@ -1,586 +1,599 @@
...
+  constructor(
+    public readonly element: UxSelectElement,
+    private styleEngine: StyleEngine,
+    private observerLocator: ObserverLocator,
+    private taskQueue: TaskQueue,
+    private defaultConfiguration: UxDefaultSelectConfiguration
+  ) {
+    // Only chrome persist the element prototype when cloning with clone node
+    Object.setPrototypeOf(element, UxSelectElementProto);
+    if (this.defaultConfiguration.theme !== undefined) {
```
bigopon, line 117: @ben-girardet just for your awareness, we now have default configuration as well.

packages/select/src/ux-select.ts
```diff
@@ -1,586 +1,599 @@
...
+  // Populated by Aurelia
+  public readonly optionWrapperEl: HTMLElement;
+  public readonly optionCtEl: UxOptionContainer;
+
+  constructor(
+    public readonly element: UxSelectElement,
+    private styleEngine: StyleEngine,
+    private observerLocator: ObserverLocator,
+    private taskQueue: TaskQueue,
+    private defaultConfiguration: UxDefaultSelectConfiguration
```
bigopon, line 113: is this configuration used anywhere after the constructor? If not, maybe let's not make it a property of the select?

MaximBalaganskiy: done

PR #289, opened 2020-06-02
chore(all): small fixes

bigopon:


packages/input/src/ux-input.html
```diff
@@ -4,10 +4,9 @@
   readonly.bind="readonly & booleanAttr"
   aria-disabled.bind="disabled & booleanAttr"
   aria-readonly.bind="readonly & booleanAttr"
-  click.trigger="focus()">
+  click.trigger="focus() || true">
```
bigopon, line 7: @MaximBalaganskiy i moved return true here so that our vm `focus()` method can stay aligned with common expectation with that method

PR #293, opened 2020-06-07
fix(lookup): hide transparent list

MaximBalaganskiy:
Also fix closing animation.

Fixes a bug when the list is invisible but still clickable

### DIFF
--- a/packages/lookup/src/ux-lookup-theme.ts
+++ b/packages/lookup/src/ux-lookup-theme.ts
@@ -6,8 +6,9 @@ export class UxLookupTheme implements UxTheme {
   static DEFAULT_INPUT_DISTANCE = 3;
   static DEFAULT_WINDOW_EDGE_DISTANCE = 15;
   static DEFAULT_BOTTOM_HEIGHT_THRESHOLD = 100;
+  static DEFAULT_TRANSITION_DURATION = '125ms';
 
-  transitionDuration: string = '125ms';
+  transitionDuration: string = UxLookupTheme.DEFAULT_TRANSITION_DURATION;
   background: string = '#F5F5F5';
   foreground: string = '#212121';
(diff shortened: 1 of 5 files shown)


packages/lookup/src/ux-lookup-theme.ts
```diff
@@ -6,8 +6,9 @@ export class UxLookupTheme implements UxTheme {
   static DEFAULT_INPUT_DISTANCE = 3;
   static DEFAULT_WINDOW_EDGE_DISTANCE = 15;
   static DEFAULT_BOTTOM_HEIGHT_THRESHOLD = 100;
+  static DEFAULT_TRANSITION_DURATION = '125ms';
```
bigopon, line 9: It's a bit weird when everything else is a number and this is a string. But maybe we can re-organize all these together at once.

bigopon:
is this ready to go?

MaximBalaganskiy:
As far as I can see, yes

packages/lookup/src/ux-lookup-theme.ts
ben-girardet: Jump in here for general comments about defaults:

I think we should avoid `static` if possible. Would probably be better to set defaults value all through the default configuration passed to the plugin. It's more clear if there is only one way to set defaults and everything goes through it.

Also, CSS vars already have defaults in the `.css` file. In the case of this lookup component it's all perfectly set from `.css`. The tendency in other components is to leave the theme empty and defaults the value direct from CSS. When we need to set defaults we should use the styleEngine and apply a theme to the body (or we can shortcut this process with a `theme` property in the default configuration object).

MaximBalaganskiy: If the theme is not set there won't be css variables. But the code needs these defaults somewhere, either as constants or static

PR #294, opened 2020-06-07
fix(select): apply initial theme value, fix transition

MaximBalaganskiy:
The component might be themed globally in which case `theme` field won't have the actual value.
Plus, number does not work with CSS, it needs to have 'ms'.

### DIFF
--- a/packages/select/src/ux-select.ts
+++ b/packages/select/src/ux-select.ts
@@ -144,6 +144,7 @@ export class UxSelect implements UxInputComponent {
     if (!this.winEvents) {
       this.winEvents = new ElementEvents(window);
     }
+    this.themeChanged(this.theme);
     // Initially Synchronize options with value of this element
     this.taskQueue.queueMicroTask(this);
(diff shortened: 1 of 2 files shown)


packages/select/src/ux-select.ts
```diff
@@ -355,13 +356,16 @@ export class UxSelect implements UxInputComponent {
     }
     this.isCollapsing = true;
     this.optionCtEl.classList.remove('ux-select__list-container--open');
+    const listTransitionString = getComputedStyle(this.element).getPropertyValue('--aurelia-ux--select-list-transition')
+      || UxSelectTheme.DEFAULT_LIST_TRANSITION;
+    const listTransition = parseInt(listTransitionString.replace('ms', ''));
```
bigopon, line 361: If we are using `parseInt`, probably no need to do a replace here, since it's at the end of the string. But maybe need to look at the config value normalization story a bit more across components for transition.

PR #296, opened 2020-06-09
fix(positioning): copy assets on build

MaximBalaganskiy:


bigopon:
doh, thanks @MaximBalaganskiy

PR #297, opened 2020-06-09
fix(modal): set max-height for desktop modals

ben-girardet:


bigopon:
I remember iOS has bug related to vh. Will this work there?

ben-girardet:
I did not notice any issue on iOS with vh for a long time. Do you have more info about this bug ?


https://caniuse.com/#feat=mdn-css_types_length_vh

ben-girardet:
I imagine you talk about this : https://stackoverflow.com/a/37113430

Conclusions: 
- it’s an intentional decision from Apple
- only affect 100vh
- in this particular case the vh is used for desktop styling

But thanks for the question, I discovered something very interesting about vh !

PR #301, opened 2020-06-12
fix(publish): update peer dependencies before publish

MaximBalaganskiy:
Lerna ignores `peerDependencies` completely whilst they need to be in line with `devDependencies`.

Also, add missing modules to `components`


bigopon:
@EisenbergEffect at the moment, dependencies are specified as peer deps. This is to combat an issue where users would have duplicated version of a modules. Though this makes lerna skip updating dependencies of packages, as it ignore peers section.

Do you see any issues with hooking into `prepublishOnly`? Description:
> prepublishOnly: Run BEFORE the package is prepared and packed, ONLY on npm publish. (See below.)

EisenbergEffect:
No objections from me.

bigopon:
@MaximBalaganskiy is there any other hook that let us have "what you see in git is what you get in npm"?

MaximBalaganskiy:
`version` hook suits better - it's run right before the commit

bigopon:
The description of `version` commands from https://docs.npmjs.com/misc/scripts. If we want t enhance this more, maybe use some more from lerna and https://docs.npmjs.com/cli/version.html
```
preversion: Run BEFORE bumping the package version.
version: Run AFTER bumping the package version, but BEFORE commit.
postversion: Run AFTER bumping the package version, and AFTER commit.
```
Looks suitable for this to me. Thanks @MaximBalaganskiy

packages/update-peers.js
```diff
@@ -0,0 +1,11 @@
+const fs = require ('fs');
+
+const data = fs.readFileSync('package.json');
+const content = JSON.parse(data.toString());
+content.peerDependencies = content.peerDependencies || {};
+for (let p of Object.getOwnPropertyNames(content.devDependencies)) {
+  if (p.startsWith('@aurelia-ux')) {
+    content.peerDependencies[p] = content.devDependencies[p];
```
bigopon, line 8: @MaximBalaganskiy are we sure that after pumping the version of a package, lerna also update the dev dependencies versions? or it waits until every package is done before bumping this? If it's the later, we may have issue here

MaximBalaganskiy: Lerna does packages one by one calling appropriate hooks. I've tried this in my local branch, works as intended.

bigopon: Ok that's awesome. Thanks for this PR. let's go then

Issue #299, opened 2020-06-09
Add aurelia-typed-observable-plugin dependency


MaximBalaganskiy:
It would be great to use `aurelia-typed-observable-plugin` in UX components.
At the moment, whenever we need a boolean or a number property, a lot of plumbing code is needed to convert strings to typed values. 

E.g.
```ts
  debounceNumber: number | null | undefined = this.defaultConfiguration.debounce;
  @bindable
  debounce: number | string | undefined = this.defaultConfiguration.debounce;
  debounceChanged() {
    this.debounceNumber = normalizeNumberAttribute(this.debounce);
  }
```
As you can see, far from convenient. Plus, I need to remember to use a shadow property instead of a bindable directly.

With the plugin this becomes just
```ts
  @bindable // this is plugin's bindable
  debounce: number = this.defaultConfiguration.debounce;
```
The advantage is obvious

EisenbergEffect:
I'd prefer to add this to the core and do it in a way compatible with Aurelia 2. @bigopon Is that doable?

bigopon:
It's doable, all the patches needed are here https://github.com/aurelia-contrib/aurelia-typed-observable-plugin/blob/master/src/patches.ts

Though aren't we constraining ourself from adding new features to v1? Also, at the moment, v2 `coerce` is `set:
```ts
@bindable({ set: v => v })
```
but the plugin mentioned is `coerce`:
```ts
@bindable({ coerce: v => v })
```
Yeah, it's probably mainly about releasing feature to v1, also the ramification of this as well, as Jeremy commented elsewhere. I think it's better in the form of a plugin. Can make it official part of ux if needed.

EisenbergEffect:
What if we move the plugin into ux but renamed it so that the APIs matched the proposed behavior of v2?

bigopon:
I can do that, but probably it will be a new package, as it doesn't seem to be simply "move" a plugin over a monorepo. Sounds like you are happy with that, I'll proceed that way

bigopon:
And tbh, that plugin matches the theme `ux` too 😄

EisenbergEffect:
Why not put it right in the core ux package?

bigopon:
hmm yeah that sounds better. Thanks 👍

bigopon:
On a 2nd thought, doing it in UX will make ux incompatible with that plugin in user app. How should we handle that situation?

MaximBalaganskiy:
Why would it be incompatible?

[...]

PR #302, opened 2020-06-15
feat(pagination): add pagination component

### LINKED ISSUE #300: Consider pagination component
I've got a pagination component which might be useful.
Please consider adding it to the UX.

pagination

MaximBalaganskiy:
uses typed plugin as agreed in #299 
fixes #300


packages/pagination/src/ux-pagination.ts
```diff
@@ -0,0 +1,83 @@
+import { customElement, useView, PLATFORM, bindingMode } from 'aurelia-framework';
+import { bindable } from 'aurelia-typed-observable-plugin';
```
bigopon, line 2: @EisenbergEffect is it ok to go ahead?

EisenbergEffect: I would prefer to resolve the design issue with au2 and then bake the same design into au1 core.

MaximBalaganskiy: @bigopon What needs to be done to progress this?

bigopon: Need to settle on how we would want to API to be for type declaration shortcut:
```ts
@bindable({ mode: 'toggle' })
@bindable({ type: 'bool' })
@bindable({ coerce: 'bool' })
@bindable({ set: 'bool' })
```

bigopon: I think for now, we can settle first with `set` api supporting only a setter function, as it's already in v2. For more high level api like `type/mode/coerce`, we can enhance it later. Does that sound good to start adding it to v1?

bigopon: @EisenbergEffect any comment on the above ☝️

PR #303, opened 2020-06-15
fix(select): move dropdown to body

MaximBalaganskiy:
This fixes an issue with parents which might have hidden overflow

Also, clean up lookup animation

### DIFF
--- a/packages/select/src/ux-select.ts
+++ b/packages/select/src/ux-select.ts
@@ -164,7 +164,7 @@ export class UxSelect implements UxInputComponent {
   }
 
   public detached() {
-    this.element.appendChild(this.optionWrapperEl);
+    this.unsetupListAnchor();
   }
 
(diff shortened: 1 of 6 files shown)


packages/select/src/ux-select.ts
```diff
@@ -163,6 +163,10 @@ export class UxSelect implements UxInputComponent {
     );
   }
 
+  public detached() {
+    this.element.appendChild(this.optionWrapperEl);
```
bigopon, line 167: let's extract this method and the one that move the `optionWrapperEl` into 2 methods: 1 to move to body and 1 to move back inside this component.

MaximBalaganskiy: done. I've also unsubscribed on `detached`

MaximBalaganskiy:
Anything else to fix here?

PR #305, opened 2020-06-17
fix(checkbox): enable label click

MaximBalaganskiy:


bigopon:
Can you give some description about the motivation for this change?

MaximBalaganskiy:
Default browser checkbox allows toggle via label clicking.
It's quite inconvenient to aim for the square which is relatively small.

bigopon:
Shouldn't connecting a label to a checkbox be the job of the user? There' ways to do it: wrapping inside a label, id + for, adding the listener by users themselves.
Also, it doesn't seem correct to only take care of immediate sibling, and especially only next sibling.

MaximBalaganskiy:
Wrapping in label works. Although styling is needed to align text with the checkbox and enable cursor pointer

PR #308, opened 2020-06-23
feat(popup): add popup

### LINKED ISSUE #307: A dropdown component
I've got a dropdown component which works similar to lookup.
It has also got default styles for items but not limited to them and can display random content.

I know this component is already being worked on.
Is it worth submitting a PR?

MaximBalaganskiy:
Fixes #307

### DIFF
--- a/packages/core/src/styles/style-engine.ts
+++ b/packages/core/src/styles/style-engine.ts
@@ -43,16 +43,4 @@ export class StyleEngine {
   public getDefaultTheme(key: string) {
     return this.styleController.themes[key as any];
   }
-
-  /**
-   * Retrieves the computed CSS variable value for the given element and key.
-   * 
-   * @param element 
-   * @param key Key of the theme
-   * @param variableName Name of the theme variable to retrieve
-   * @param defaultValue Default value
-   */
-  public getVariableValue(element: Element, key: string, variableName: string, defaultValue?: string): string {
-    return getComputedStyle(element).getPropertyValue(`--aurelia-ux--${key}-${variableName}`) || defaultValue || '';
-  }
 }
--- a/packages/core/src/styles/style-engine.ts
+++ b/packages/core/src/styles/style-engine.ts
@@ -43,4 +43,16 @@ export class StyleEngine {
   public getDefaultTheme(key: string) {
     return this.styleController.themes[key as any];
   }
+
+  /**
+   * Retrieves the computed CSS variable value for the given element and key.
+   * 
+   * @param element 
+   * @param key Key of the theme
+   * @param variableName Name of the theme variable to retrieve
+   * @param defaultValue Default value
+   */
+  public getVariableValue(element: Element, key: string, variableName: string, defaultValue?: string): string {
+    return getComputedStyle(element).getPropertyValue(`--aurelia-ux--${key}-${variableName}`) || defaultValue || '';
+  }
 }
--- a/packages/popup/src/ux-popup.ts
+++ b/packages/popup/src/ux-popup.ts
@@ -26,10 +26,16 @@ export class UxPopup implements UxComponent, EventListenerObject {
   public isMeasured: boolean = false;
 
   @bindable
-  trigger: HTMLElement;
-  triggerChanged(newValue: HTMLElement, oldValue: HTMLElement) {
-    oldValue?.removeEventListener('click', this);
-    newValue?.addEventListener('click', this);
+  trigger: HTMLElement | string | null;
+  triggerChanged(newValue: HTMLElement | string, oldValue: HTMLElement | string) {
+    if (typeof (newValue) === 'string') {
+      this.trigger = document.querySelector<HTMLElement>(newValue);
+    } else {
+      if (oldValue && typeof (oldValue) !== 'string') {
+        oldValue.removeEventListener('click', this);
+      }
+      newValue?.addEventListener('click', this);
+    }
   }
 
   @bindable
@@ -46,7 +52,9 @@ export class UxPopup implements UxComponent, EventListenerObject {
   autoclose: string | boolean = true;
 
   detached() {
-    this.trigger?.removeEventListener('click', this);
+    if (this.trigger && typeof (this.trigger) !== 'string') {
+      this.trigger?.removeEventListener('click', this);
+    }
   }
 
   handleEvent(evt: Event): void {
@@ -91,7 +99,7 @@ export class UxPopup implements UxComponent, EventListenerObject {
   }
 
   updateAnchor() {
-    if (!this.trigger) {
+    if (!this.trigger || typeof (this.trigger) === 'string') {
       return;
     }
     const rect = this.trigger.getBoundingClientRect();
--- a/packages/popup/src/ux-popup.ts
+++ b/packages/popup/src/ux-popup.ts
@@ -146,21 +146,14 @@ export class UxPopup implements UxComponent, EventListenerObject {
   }
 
   onWindowClick(evt: Event) {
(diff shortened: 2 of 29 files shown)


packages/core/src/styles/style-engine.ts
```diff
@@ -43,4 +43,16 @@ export class StyleEngine {
...
+
+  /**
+   * Retrieves the computed CSS variable value for the given element and key.
+   * 
+   * @param element 
+   * @param key Key of the theme
+   * @param variableName Name of the theme variable to retrieve
+   * @param defaultValue Default value
+   */
+  public getVariableValue(element: Element, key: string, variableName: string, defaultValue?: string): string {
```
bigopon, line 55: can we keep this private to each package for now? we can always extract it later.

packages/popup/src/ux-popup.ts
```diff
@@ -0,0 +1,157 @@
...
+  constructor(private element: HTMLElement, private styleEngine: StyleEngine, private taskQueue: TaskQueue) { }
+
+  public anchor: IAnchor | null;
+  public isWrapperOpen: boolean = false;
+  public isOpen: boolean = false;
+  public popupContent: HTMLDivElement;
+  public isMeasured: boolean = false;
+
+  @bindable
+  trigger: HTMLElement;
```
bigopon, line 29: I think it's also reasonable to support the usage:
```html
<ux-popup trigger="#someid">
```
For that, we may need to tweak this trigger abit

packages/popup/src/ux-popup.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  onWindowClick(evt: Event) {
+    if (!this.isOpen || !normalizeBooleanAttribute('autoclose', this.autoclose)) {
+      return;
+    }
+    let triggerClicked = false;
+    let parent: HTMLElement | null = evt.target as HTMLElement;
+    while (parent) {
```
bigopon, line 146: should we do a `evt.target.contains(this.trigger)` ? it seems simpler to me

packages/core/src/styles/style-engine.ts
MaximBalaganskiy: done

packages/popup/src/ux-popup.ts
MaximBalaganskiy: done

packages/popup/src/ux-popup.ts
MaximBalaganskiy: done

packages/popup/src/ux-popup.ts
```diff
@@ -0,0 +1,170 @@
...
+  public isWrapperOpen: boolean = false;
+  public isOpen: boolean = false;
+  public popupContent: HTMLDivElement;
+  public isMeasured: boolean = false;
+
+  @bindable
+  trigger: HTMLElement | string | null;
+  triggerChanged(newValue: HTMLElement | string, oldValue: HTMLElement | string) {
+    if (typeof (newValue) === 'string') {
+      this.taskQueue.queueTask(() => this.trigger = DOM.querySelector(newValue) as HTMLElement);
```
bigopon, line 32: I think this will make the functionality a bit brittle, any trigger must be added before this popup, otherwise it's gonna fail. I think we will need to keep it as is if it's a string, and perform a single check inside `onWindowClick`

MaximBalaganskiy:
We need to subscribe to triggers click as well

bigopon:
We don't need to if we check for the click target in `onWindowClick`

MaximBalaganskiy:
here you go... but this will always listen to window events as opposed to the previous iteration when the element only subscribed on open

MaximBalaganskiy:
I've discovered an unpleasant side effect when autoclose is false - changing previous elements height moves the trigger away from the popup. It is something we should be addressing?

packages/popup/src/ux-popup.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  onWindowClick(evt: Event) {
+    if (typeof (this.trigger) === 'string') {
+      this.trigger = DOM.querySelector(this.trigger) as HTMLElement;
+    }
+    if (this.trigger && this.trigger.contains(evt.target as HTMLElement)) {
+      this.triggerClick();
+    } else if (this.isOpen && normalizeBooleanAttribute('autoclose', this.autoclose)) {
+      this.close();
+    }
```
bigopon, line 146: I think we can avoid mutation of `trigger`, since there's no need for it

```suggestion
    const trigger= typeof this.trigger === 'string'
      ? DOM.querySelector(this.trigger)
      : this.trigger;
    if (trigger != null && trigger.contains(evt.target as HTMLEelement)) {
      this.triggerClick();
    } else if (this.isOpen && normalizeBooleanAttribute('autoclose', this.autoclose)) {
      this.close();
    }
```

bigopon:
> here you go... but this will always listen to window events as opposed to the previous iteration when the element only subscribed on open

Well, yes, it does sound bad. Maybe we can have a single listener for all popups. But this should be prioritized over the other way, as we likely get complains from brittle API.

> I've discovered an unpleasant side effect when autoclose is false - changing previous elements height moves the trigger away from the popup. It is something we should be addressing?

There's no easy way to handle this I think. Something like this would probably be called position-observer. And it needs to observe the position of an element, not only when user scrolls, resizes windows, but also when CSS classes/inline styles changes. We can offer a simple implementation to deal with scroll/resize, but maybe it's not that good. I'm thinking let each popup listen to a specific event on document, and based on that to update the anchor accordingly. So the consumer of a popup can have a way to decouple the position update. Something like this:

```ts
interface IPopupTriggerPositioningEventDetail {
  trigger: HTMLElement;
  position: DOMRect;
}

// in application code:
document.dispatchEvent(new CustomEvent('popup-trigger-positioning', { details: someButton }));
```

MaximBalaganskiy:
If only we could use absolute position :)

MaximBalaganskiy:
Not sure what complains you mean. For a consumer it doesn't matter how we implemented it. Do you see issues with using task queue?

MaximBalaganskiy:
We could also change the API and expose an `open` method which accepts an anchor element reference. Although this will be less convenient to use.

MaximBalaganskiy:
Tbh, I would just bind to a true trigger reference. What's the advantage of using a query string?

MaximBalaganskiy:
So, what do we do here? I vote for removing string selector. It provides no benefit compared to `ref`

bigopon:
Yeah probably. Lets start simple

MaximBalaganskiy:
I've removed it

packages/popup/src/ux-popup.ts
```diff
@@ -0,0 +1,160 @@
...
+  public anchor: IAnchor | null;
+  public isWrapperOpen: boolean = false;
+  public isOpen: boolean = false;
+  public popupContent: HTMLDivElement;
+  public isMeasured: boolean = false;
+
+  @bindable
+  trigger: HTMLElement | undefined;
+  triggerChanged(newValue: HTMLElement, oldValue: HTMLElement) {
+    oldValue && oldValue.removeEventListener('click', this);
```
bigopon, line 31: Well we need to be more consistent in how we use operators. Also we should be aiming for less clever code

bigopon:
Looks good to me as the first iteration. Thanks @MaximBalaganskiy

packages/popup/src/ux-popup.ts
MaximBalaganskiy: Would be nice to use ?. which fails tests...

PR #309, opened 2020-06-24
feat(lookup): allow preloading options

MaximBalaganskiy:
it's a useful feature when a number of options is limited and a user wants to see them before filtering

### DIFF
--- a/packages/lookup/src/ux-lookup.ts
+++ b/packages/lookup/src/ux-lookup.ts
@@ -137,7 +137,7 @@ export class UxLookup implements UxComponent, EventListenerObject {
     }
     lookupEvents.forEach(x => this.element.addEventListener(x, this));
     this.valueChanged();
-    if (normalizeBooleanAttribute('preload-options', this.preloadOptions)) {
+    if (!this.value && normalizeBooleanAttribute('preload-options', this.preloadOptions)) {
       try {
         await this.loadOptions();
       }
--- a/packages/lookup/src/ux-lookup.ts
+++ b/packages/lookup/src/ux-lookup.ts
@@ -130,7 +130,7 @@ export class UxLookup implements UxComponent, EventListenerObject {
     this.optionsChanged();
   }
 
-  async attached() {
+  attached() {
     this.inputElement = this.element.parentElement?.querySelector<UxInputElement | HTMLInputElement>('ux-input,input');
     if (this.inputElement) {
       inputEvents.forEach(x => this.inputElement!.addEventListener(x, this));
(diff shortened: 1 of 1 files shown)


bigopon:
I think some higher level features like this should at least be delayed to a later phase, so we can avoid expanding the core with features that are not fundamental.

MaximBalaganskiy:
I think this is quite necessary feature though. The component is extremely inconvenient to use without it.

packages/lookup/src/ux-lookup.ts
```diff
@@ -120,20 +120,29 @@ export class UxLookup implements UxComponent, EventListenerObject {
...
+  async attached() {
     this.inputElement = this.element.parentElement?.querySelector<UxInputElement | HTMLInputElement>('ux-input,input');
     if (this.inputElement) {
       inputEvents.forEach(x => this.inputElement!.addEventListener(x, this));
     }
     lookupEvents.forEach(x => this.element.addEventListener(x, this));
     this.valueChanged();
+    if (!this.value && normalizeBooleanAttribute('preload-options', this.preloadOptions)) {
+      try {
+        await this.loadOptions();
```
bigopon, line 142: why do we need await here?

bigopon:
If we are to support a loading feature, should it be a way to resolve all kind of promises so that during loading, it would display something and upon resolving the async value, would display that?

packages/lookup/src/ux-lookup.ts
MaximBalaganskiy: No particular reason other than being async all the through

MaximBalaganskiy:
Can you give an example please?

packages/lookup/src/ux-lookup.ts
bigopon: I see. Maybe let's avoid it if we don't need it. Having await in `attached` and await here generates bloated code as well.

bigopon:
From the changes, it doesn't seem to load anything, can you help give an example of how it's supposed to be used?

MaximBalaganskiy:
Let's say you just landed on a page with a lookup. If you just click on a lookup the popup won't have anything until you start typing. Preloading options would let you see them immediately on the first click. Useful when you don't have what you're looking for.

bigopon:
That makes sense, though do we need to wait until `attached` to preload? it seems it can start during bind already

packages/lookup/src/ux-lookup.ts
MaximBalaganskiy: removed

MaximBalaganskiy:
I just wanted to avoid calling all the changed handlers manually

PR #311, opened 2020-06-26
feat(expandable): allow specifying an accordion group id

MaximBalaganskiy:
Not sure why I had it so complex in the first place.

Plus allow specifying an accordion group id,

### DIFF
--- a/packages/expandable/src/ux-expandable.ts
+++ b/packages/expandable/src/ux-expandable.ts
@@ -6,7 +6,7 @@ import { UxExpandableTheme } from './ux-expandable-theme';
 @customElement('ux-expandable')
 @useView(PLATFORM.moduleName('./ux-expandable.html'))
 export class UxExpandable implements UxComponent {
-  constructor(public element: HTMLElement, private styleEngine: StyleEngine) { }
+  constructor(public element: HTMLElement, private styleEngine: StyleEngine, private taskQueue: TaskQueue) { }
 
   static OPEN_CHANGED_EVENT = 'open-changed';
 
@@ -39,6 +39,20 @@ export class UxExpandable implements UxComponent {
   @bindable
   accordion: string | undefined = undefined;
 
+  handleEvent(e: Event) {
+    switch (e.type) {
+      case 'transitionend':
+        this.setContentContainerHeightToAuto();
+        break;
+    }
+  }
+
+  setContentContainerHeightToAuto() {
+    this.contentContainer.style.overflow = "visible";
+    this.contentContainer.style.height = "auto";
+    this.contentContainer.removeEventListener("transitionend", this);
+  }
+
   bind() { }
 
   attached() {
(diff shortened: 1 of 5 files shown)


packages/expandable/src/ux-expandable.ts
```diff
@@ -61,22 +47,17 @@ export class UxExpandable implements UxComponent {
...
+      this.contentContainer.style.height = '0';
     }
   }
 
   toggle() {
-    if (!this.openBoolean && normalizeBooleanAttribute('accordion', this.accordion)) {
-      const otherAccordions = Array.from(this.element!.parentElement!.querySelectorAll('ux-expandable[accordion].ux-expandable--open'));
+    if (!this.openBoolean && this.accordion !== undefined) {
+      const otherAccordions = this.accordion === ''
+        ? Array.from(this.element!.parentElement!.querySelectorAll('ux-expandable[accordion].ux-expandable--open'))
```
bigopon, line 59: I just realized that using `.querySelectorAll` will select all expandable. Even the ones nested inside this, is this the intention?

packages/expandable/src/ux-expandable.ts
```diff
@@ -61,22 +47,17 @@ export class UxExpandable implements UxComponent {
...
-      this.contentContainer.style.height = this.content.clientHeight + "px";
+      this.contentContainer.style.height = this.content.clientHeight + 'px';
     } else {
-      // the following line is needed because height has been restored to auto"
-      this.contentContainer.style.height = this.content.clientHeight + "px";
-      this.taskQueue.queueTask(() => {
-        this.contentContainer.style.overflow = "hidden";
-        this.contentContainer.style.height = "0";
-      });
+      this.contentContainer.style.height = '0';
```
bigopon, line 52: Doesn't it need to keep the hidden overflow to make sure its content is hidden.

MaximBalaganskiy: It's hidden in the css

packages/expandable/src/ux-expandable.ts
MaximBalaganskiy: Not even sure if nested expandables would work. So no, there was no such intent

packages/expandable/src/ux-expandable.ts
bigopon: nice, thanks. I just realized we may need to observe the size of the content and adjust the size of this container accordingly, otherwise when the content grows, it will be clipped. or we can wait until the transition finishes and remove the fixed height

MaximBalaganskiy: Oh, that might be the reason I had it so complex in the first place... To support growing content

MaximBalaganskiy: I've reverted this change

packages/expandable/src/ux-expandable.ts
MaximBalaganskiy: Children expandables work fine. To prevent children from collapsing one has to set `accordion` attribute to some unique value on a group of collapsibles.

MaximBalaganskiy:
Renamed the PR

PR #321, opened 2020-09-01
Fix(datepicker): respect locale specific order for weekdays

ben-girardet:
Currently the weekdays in the calendar always start on Sunday. However this is a locale specific thing and we should call `weekdays(true)` in order to follow locale in used by moment.

See https://stackoverflow.com/questions/43520233 for more infos.


bigopon:
Thanks @ben-girardet 👍 
@EisenbergEffect queuing for a release of this, together with ui-virtualization

ben-girardet:
Thanks for the quick merge !

PR #326, opened 2020-09-23
fix(input): disabled styles for input component

ben-girardet:
When an input field has no value, without this fix the input textbox is displayed with a background color and overflow the ux-input wrapper.

This fix hide the input textbox for disabled empty fields, leaving only the label in position.

Naturally the textbox is displayed for non-empty fields.


bigopon:
Thanks @ben-girardet

